Issue #81, opened 2026-02-05
extension: SAML


nacx:
Implement an extension that leverages SAML.

PR #197, opened 2026-02-24
extensions: add lazy IdP metadata fetch via Envoy HttpCallout to SAML extension

sergicastro:
Add support for fetching IdP metadata from a URL at runtime instead of requiring it inline or as a file at config time.

The metadata is fetched lazily on the first request using Envoy's HttpCallout mechanism and cached for all subsequent requests. Concurrent requests during the fetch are paused and resumed once metadata is available, with the full SAML auth flow (session check, IdP redirect) running explicitly after the fetch since OnRequestHeaders is not re-invoked after ContinueRequest.

Follow up #81


wbpcode (commented):
Thanks for this contribution. I think the core problem here is we didn't support http callout in the config handle, right? Or we actually could load the configuration from remote when loading the configuration?
Current implementation is more like a compromised choice because the limited SDK/ABI and are a little tricky.

Because this is more like an improvement rather than block feature, may be we can defer this, and I will try to back port/enhance the SDK then we can have a better implementation?

WDYT?

---

(Another way is to use the http client of golang self to load the resource when loading the configuration (it's one time operation, I don't think it require connection pool/outlier detection and so on. So, although it's not suggested, I think still acceptable to use that)

nacx:
@sergicastro @wbpcode once https://github.com/tetratelabs/built-on-envoy/pull/387 is merged, this should be unblocked

sergicastro:
I still have this on my radar.
I see the pr already merged.
I'll find time this week.

sergicastro:
Not ready for review, there's an issue when performing `httpCallout` at config state because of clusters not yet ready.
I will work on a fix.

sergicastro:
Got the issue fixed by scheduling the IdP fetch to give time to clusters to be available.
Thanks @wbpcode for the suggestion and help.

extensions/composer/saml/plugin.go
```diff
@@ -102,16 +124,153 @@ func (f *HTTPFilterConfigFactory) Create(handle shared.HttpFilterConfigHandle, c
...
+		{":authority", u.Host},
+	}
+
+	handle.Log(shared.LogLevelInfo,
+		"saml: idp metadata fetch scheduled from %s via cluster %s after %s",
+		cfg.IDPMetadataURL, cfg.IDPMetadataCluster, cfg.IDPMetadataFetchDelay)
+
+	go func() {
+		time.Sleep(cfg.IDPMetadataFetchDelay)
+		scheduler.Schedule(func() {
```
nacx: This is a one attempt after the fetch delay, right? Is it worth doing some retry attempts here instead of only relying on the startup value?

sergicastro: We could.
But in case we want to add retries with wait or backoff, in dynamic modules, we don't have the retry API, meaning we should implement something similar, like adding a loop in this go routine managing the sleeps and backoff.

I can work on this if you think it's worth it.

nacx: I don't think we need to get sophisticated here, but just a fixed retry of, say, 3 attempts, separated by the configured fetch delay, is enough.
The main reason is that the causes for a slow startup may vary, and if the only option to overcome that is to reconfigure and reload the config, that's probably not the best. Having some basic retry to maximize success withotu manual intervention would be great.

sergicastro: done

nacx (approved):

PR #321, opened 2026-04-07
Enhance llm-proxy with richer LLM observability

Aias00:
## Summary
- enhance `llm-proxy` with richer observability instead of introducing a separate overlapping plugin
- add richer request/response metadata for OpenAI Chat Completions and Anthropic Messages, including prompts, answers, reasoning, tool calls, and token detail fields
- add optional structured log output, session ID extraction, and keep streaming/non-streaming observability behavior aligned
- preserve existing custom matcher semantics while allowing built-in default OpenAI/Anthropic paths to match when query strings are present

## Verification
- `cd extensions/composer && go test ./llm-proxy/...`
- `cd extensions/composer && go test ./...`
- `make -C cli gen`
- `cd website && npm run build`

## Notes
- this branch intentionally avoids a separate `llm-statistics` plugin and consolidates the richer observability behavior into `llm-proxy`
- `cli/e2e` Docker-dependent paths were not exercised in this branch


wbpcode (commented):
Thanks for this update. I guess it's a good requirement for debug/observe AI traffic.

But note the current metadata could only support number/bool/string and cannot support other typed data. I personally prefer to avoid metadata to take raw payload for now because it's inefficient.

We are working a new abstraction which will expose more details of traffic and also could support auto-transformation, I think will also helpful for your requirements. I guess we may could wait that PR first and I will also ping your suggestions there.

Feel free to slack me at Envoy Slack or BoE slack (with name wbpcode) because I think you are Chinese (?) and have same timezone with me.

PR #353, opened 2026-04-09
cli: add per-route-config to scaffolding and rust extensions

nacx:


extensions/ip-restriction/src/lib.rs
```diff
@@ -115,6 +133,18 @@ impl<EHF: EnvoyHttpFilter> HttpFilter<EHF> for Filter {
...
+        let effective_config =
+            if let Some(route_cfg) = envoy_filter.get_most_specific_route_config() {
+                if let Some(per_route) = route_cfg.downcast_ref::<PerRouteFilterConfig>() {
+                    per_route.config.clone()
+                } else {
+                    self.filter_config.config.clone()
+                }
+            } else {
+                self.filter_config.config.clone()
+            };
```
wbpcode: I guess it would be better to evaluate the route level configuration at new_http_filter.

By this way, the `Filter` self needn't to care configuration overriding and more friendly for test, easy to understand and keep the consistency with other BoE filters (initial matched's route's configuration works)

wbpcode (commented):
Thanks for providing this. I guess you may need to run `cargo clippy` locally?

extensions/ip-restriction/src/lib.rs
nacx: Done

cli/cmd/templates/create/rust/lib.rs.tmpl
```diff
@@ -46,16 +47,49 @@ impl FilterConfig {
...
+        } else {
+            self.header_value.clone()
+        };
+        Box::new(Filter { header_value })
     }
 }
 
 pub struct Filter {
-    filter_config: FilterConfig,
+    header_value: String,
```
wbpcode: ? Could you re-check this?

nacx: Since there was only one property I put it here, but I agree that it may be better to have a config object here inside the filter. I'll update.

wbpcode: I guess we should parse the route configuration to a `FilterConfigImpl` and then could use it for ip restrictions?

nacx: I've updated the code to follow the same pattern we were using in the ip-restriction

wbpcode (approved):

PR #354, opened 2026-04-10
extension: bump dynamic_modules SDK to pick up body buffer fixes

codefromthecrypt:
Pin github.com/envoyproxy/envoy/source/extensions/dynamic_modules to v0.0.0-20260410020555-b461f159615c which fixes FakeBodyBuffer.Drain and adds ReceivedBufferedRequestBody/ReceivedBufferedResponseBody methods.

Remove testBodyBuffer workarounds that were compensating for the Drain bug and add mock expectations for the new bool methods.

### DIFF
--- a/extensions/composer/anthropic-decoder/plugin_test.go
+++ b/extensions/composer/anthropic-decoder/plugin_test.go
@@ -14,33 +14,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
-
-	"github.com/tetratelabs/built-on-envoy/extensions/composer/pkg"
 )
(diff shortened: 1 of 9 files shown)


wbpcode:
This is dangerous because the new sdk means new ABI which is not supported in old Envoy.

Let's the CI tell us could that works.

wbpcode:
Sure it can't work because the 1.37.1 doesn't provide the necessary ABI that ReceivedBufferedResponseBody required.

I will back port the SDK fix to 1.37 and then we can update the SDK to latest version of Envoy 1.37

codefromthecrypt:
Ah cool. Yeah ABI memories in wasm returning;)

nacx (approved):

PR #356, opened 2026-04-12
build: bump to Go 1.26.2 to fix CVEs

nacx:
There are many CVEs that are fixed in Go 1.26.2


wbpcode:
Should we also update the golang extension template here?

nacx:
Good catch. I'll do!

nacx:
@wbpcode addressed the comments. Can you take another look?

wbpcode (approved):

PR #362, opened 2026-04-13
docs: document how to use extensions in prod

nacx:
cc @sjoukedv

### DIFF
--- a/website/src/pages/docs/packaging-prod.mdx
+++ b/website/src/pages/docs/packaging-prod.mdx
@@ -68,6 +68,14 @@ ENV GODEBUG=cgocheck=0
 To configure the Dynamic Modules you can follow the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules).
 You basically need to inject the Dynamic Module filter in your HTTP filter chain.
 
+<Callout type="note" title="Required environment variables">
+The following environment variables need to be exported:
+```shell
+ENVOY_DYNAMIC_MODULES_SEARCH_PATH=/path/to/extensions  # Path where the Dynamic Module `.so` files are
+GODEBUG=cgocheck=0  # Needed to disable the cgo pointer checks as Envoy may hold pointers to Go memory
+```
+</Callout>
+
 **Rust Dynamic Module configuration example:**
 
 ```yaml
--- a/website/src/pages/docs/packaging-prod.mdx
+++ b/website/src/pages/docs/packaging-prod.mdx
@@ -33,9 +33,8 @@ boe download composer --platform linux/amd64
 ```
 
 <Callout type="tip" title="Go extensions">
-Built On Envoy provides the `composer` Dynamic Module that packages all the Go extensions, so there is no need to download Go
-extensions individually. The [Dynamic Modules and Go](#dynamic-modules-and-go) section further expands on the nuances of Go Dynamic
-Modules in Envoy.
+Built On Envoy provides the `composer` Dynamic Module that packages **all the Go extensions**. The [Dynamic Modules and Go](#dynamic-modules-and-go)
+section further expands on the nuances of Go Dynamic Modules in Envoy.
 </Callout>
 
 Once the extensions are downloaded, you need to make the `.so` files available to Envoy. This could be a shared Kubernetes
@@ -115,7 +114,7 @@ Running Go-based Dynamic Modules in Envoy has some limitations. Envoy only suppo
 using Go-based Dynamic Modules needs some careful considerations.
 
 <Callout type="danger" title="Go runtime limitations">
-Envoy only supports loading one Go runtime in-process. This means, effectively, that only one `lib<module>.so` file can be loaded by Envoy safely.
+Envoy only supports loading one Go runtime in-process. This means, effectively, that **only one Go-based `lib<module>.so` file can be loaded by Envoy safely**.
 </Callout>
 
 Built On Envoy makes it easier to overcome this limitation while providing maximum flexibility by providing:
--- a/website/src/pages/docs/packaging-prod.mdx
+++ b/website/src/pages/docs/packaging-prod.mdx
@@ -1,0 +1,152 @@
+---
+layout: ../../layouts/DocsLayout.astro
+title: Packaging and using extensions in prod
+description: Packaging the Dynamic Modules extensions to use them in Production
+---
+
+import ToC from '../../components/ToC.astro';
+import Callout from '../../components/Callout.astro';
+
+# Using the Dynamic Modules extensions in Production
+
+Built On Envoy already provides the extensions compiled as Envoy [Dynamic Modules](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules)
+for `linux/amd64` and `linux/arm64`. Extensions are packaged as OCI artifacts in the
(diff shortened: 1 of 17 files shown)


website/src/pages/docs/packaging-prod.mdx
```diff
@@ -0,0 +1,152 @@
...
+### Option 1: Download extensions manually
+
+Extensions can be easily downloaded using the [bow download](/docs/cli/download) command:
+
+```bash
+boe download ip-restriction --platform linux/amd64
+```
+
+Once the extensions are downloaded, you need to make the `.so` files available to Envoy. This could be a shared Kubernetes
+volume, a custom Envoy image, or any approach that is convenient.
```
sjoukedv, line 41: Do we need to be extra explicit here with a note about composer extensions only needing `compose.so`?

website/src/pages/docs/packaging-prod.mdx
```diff
@@ -0,0 +1,152 @@
...
+    '@type': type.googleapis.com/envoy.extensions.filters.http.dynamic_modules.v3.DynamicModuleFilter
+    dynamic_module_config:
+      name: ip-restriction
+    filter_name: ip-restriction
+    filter_config:
+      '@type': type.googleapis.com/google.protobuf.StringValue
+      value: '{"deny_addresses": ["127.0.0.1"]}'
+```
+
+**Go Dynamic Module configuration example:**
```
sjoukedv, line 92: This is the composer specific example. Can we also add an example for the standalone configuration option without composer?

As the size of composer grows with extensions this means we still have the option to only "extract" the single extension we need.

website/src/pages/docs/packaging-prod.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+```yaml
+- name: chat-completions-decoder
+  typed_config:
+    '@type': type.googleapis.com/envoy.extensions.filters.http.dynamic_modules.v3.DynamicModuleFilter
+    dynamic_module_config:
+      load_globally: true
+      name: composer
+    filter_name: chat-completions-decoder
```
sjoukedv, line 100: Maybe a note about not using the `goplugin-loader` syntax that is generated from `boe gen-config` or will it get a flag in the future with e.g. `--prod` ?

website/src/pages/docs/packaging-prod.mdx
nacx: Done

website/src/pages/docs/packaging-prod.mdx
nacx: Done

website/src/pages/docs/packaging-prod.mdx
nacx: I'd rather not add the note and have the `--prod` (or whatever better flag name) flag in the config generation commands.

website/src/pages/docs/packaging-prod.mdx
```diff
@@ -0,0 +1,167 @@
...
+
+# Setup the env vars Envoy needs to find the extensions and safely run the Go Dynamic Modules
+ENV ENVOY_DYNAMIC_MODULES_SEARCH_PATH=/var/boe/extensions
+ENV GODEBUG=cgocheck=0
+```
+
+## Configuring the Dynamic Modules
+
+To configure the Dynamic Modules you can follow the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules).
+You basically need to inject the Dynamic Module filter in your HTTP filter chain.
```
wbpcode, line 68: We may need to callout here that we should set following environment variables if it's not set by default:

```
ENV ENVOY_DYNAMIC_MODULES_SEARCH_PATH=/var/boe/extensions
ENV GODEBUG=cgocheck=0
```

In the above option 1, there no content about the environment variables.

nacx: Done!

website/src/pages/docs/packaging-prod.mdx
```diff
@@ -0,0 +1,167 @@
...
+* All loaded Go Dynamic Modules must have been compiled with the same version of Go.
+* The transitive dependencies for all loaded Go Dynamic Modules must be the same version across the Dynamic modules.
+
+<Callout type="danger" title="Runtime constraints">
+Failing to meet these constraints may result in undefined behavior when Envoy runs the Go Dynamic Modules.
+</Callout>
+
+Built On Envoy is designed to favour satisfying these constraints by providing a code structure for the extensions that can be safely used to compile
+multiple Go extensions into a single Go Dynamic Module that can be safely loaded and used in Envoy.
+
```
wbpcode: The go version and dependencies limits are not limits for go dynamic module, but for go plugin.
For the one who want to use this in production, I think we needn't to hide concept of go plugin and we need ensure our document be accurate regarding technology.

So, I guess we could organize these content in following way:

```
## Dynamic Modules and Go

> To tell only one go runtime in-process and we need to ensure there is no any other go lib
> or extension is used in the envoy process and so on. And then to tell we provided a code 
> structure for the extensions that can be safely used to compile multiple go extensions into
> a single go Dynamic Module and so on.

### Using all the Built On Envoy Go Extensions (recommended, works out-of-the-box)

...

### Building your own Dynamic Module with a subset of extensions

...

### Building your own extension as standalone go plugin (Alpha)

> To tell we designed a solution for loading custom out-of-tree plugins/extensions with
> go plugin. Then tell the go version and dependencies limits here about go plugin.

```

By this way, three options here one by one, and more and more complex. If users want the third options, they need to take the complexity and risk.

---

Just for reference and feel free to lmk your thoughts.

wbpcode:
Thanks for this great document!!!

website/src/pages/docs/packaging-prod.mdx
```diff
@@ -106,35 +105,126 @@ GODEBUG=cgocheck=0  # Needed to disable the cgo pointer checks as Envoy may hold
...
+Running Go-based Dynamic Modules in Envoy has some limitations. Envoy only supports **loading one Go runtime in-process**, and that means that
+using Go-based Dynamic Modules needs some careful considerations.
 
-* All loaded Go Dynamic Modules must have been compiled with the same version of Go.
-* The transitive dependencies for all loaded Go Dynamic Modules must be the same version across the Dynamic modules.
-
-<Callout type="danger" title="Runtime constraints">
-Failing to meet these constraints may result in undefined behavior when Envoy runs the Go Dynamic Modules.
+<Callout type="danger" title="Go runtime limitations">
+Envoy only supports loading one Go runtime in-process. This means, effectively, that only one `lib<module>.so` file can be loaded by Envoy safely.
```
wbpcode: It would be good to tell more clearly only one Go `lib<module>.so` file can be loaded by Envoy safely.

website/src/pages/docs/packaging-prod.mdx
nacx: I've reworked the whole doc, PTAL.

wbpcode (commented):
LGTM overall for now except one nit comment

website/src/pages/docs/packaging-prod.mdx
nacx: What do you mean with more clearly? :) Do you have a proposed text?

wbpcode: Sorry, I just taken a shower. But I think you did the best

wbpcode (approved):

nacx:
Thanks for hte feedback!

PR #363, opened 2026-04-15
tests(waf): more tests, reduces ignored ftw tests

M4tteoP:
follows https://github.com/tetratelabs/built-on-envoy/pull/326


nacx (approved):
Nice!

PR #366, opened 2026-04-16
extension: DNS Gateway

ovadiagal:
Adds the DNS gateway extension, which allows Envoy to allocate virtual IPs and then look up associated metadata 

diagram

In this dynamic module, you'll find: 

### `dns_gateway/mod.rs`

DNS gateway UDP listener filter. Intercepts DNS queries and returns synthetic responses for domains matching configured egress policies (exact or *. wildcard).

  - For matching A-record queries, allocates a virtual IP from the shared cache and responds with it
  - For matching non-A queries (e.g. AAAA), responds with NODATA (no IPv6 support yet)
  - For non-matching queries, passes through to the next filter

###  `virtual_ip_cache.rs`

Maintains a mapping between FQDN and virtual IPs.

  - dns_gateway calls `virtual_ip_cache::allocate()` to get a virtual IP for a domain (reuses existing IP if the domain was seen before)
  - cache_lookup calls `virtual_ip_cache::lookup()` to resolve a virtual IP back to its domain and metadata
  - IP space is bounded by the configured prefix_len; no eviction currently. 

###  `cache_lookup.rs`

Network filter that runs on new TCP connections. Looks up the destination virtual IP in the shared cache and writes the resolved domain and policy metadata into Envoy filter state, making them available to downstream filters via:
  - `%FILTER_STATE(envoy.dns_gateway.domain:PLAIN)%`
  - `%FILTER_STATE(envoy.dns_gateway.metadata.<key>:PLAIN)%`


nacx:
Can you run `make -C cli gen` ? This will regenerate the extension index and get the extensoin listed in the extension catalog. This should also trigger the netlify build for the PR to see how it looks like when rendered.

extensions/dns-gateway/manifest.yaml
```diff
@@ -0,0 +1,122 @@
...
+  - security
+  - transparent-proxy
+  - virtual-ip
+license: Apache-2.0
+type: rust
+examples:
+  - title: Single Wildcard Domain
+    description: Intercept all *.aws.com DNS queries and allocate virtual IPs from a /24 range
+    code: |
+      # dns_gateway filter config
```
nacx: In the examples, can you put the while `boe run --extension <dns-gateway>` command, so it is trivial for people that wants to try to copy/paste ?

extensions/dns-gateway/manifest.yaml
```diff
@@ -0,0 +1,122 @@
...
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+name: dns-gateway
+version: 0.1.0
+minEnvoyVersion: 1.37.1
+categories:
+  - Network
+  - Security
+author: ovadiagal
```
nacx: Can we have the full name here, if you want this to be listed as an individual? Or should this better be the name of the company?

extensions/dns-gateway/src/dns_gateway/cache_lookup.rs
```diff
@@ -0,0 +1,156 @@
...
+            .unwrap();
+
+        let config = CacheLookupFilterConfig::new(b"");
+        let mut mock = MockEnvoyNetworkFilter::new();
+        let mut filter = config.new_network_filter(&mut mock);
+
+        let mut mock = MockEnvoyNetworkFilter::new();
+        mock.expect_get_local_address()
+            .returning(move || (ip.to_string(), 8080));
+        mock.expect_set_filter_state_bytes().returning(|_, _| true);
```
nacx: Can you make an explicit assertion to verify it has been called with the right parameter values? Since downstream filters depend on this, it is important to have explicit assertions on the values to catch any regression.

extensions/dns-gateway/src/dns_gateway/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+
+            match question.query_type() {
+                RecordType::A => {
+                    let base_ip: u32 = matcher.base_ip.parse::<Ipv4Addr>().ok()?.into();
+                    let ip = get_cache().allocate(
+                        domain,
+                        matcher.metadata.clone(),
+                        base_ip,
+                        matcher.prefix_len as u8,
+                    )?;
```
nacx: IIUC, if the CIDR range is exhausted and allocation fails, then this will end up causing a `return Continue;`, forwarding that upstream. Clients will get the DNS response from the upstream and not from the proxy.

This is fine, but not aligned with other behaviours, for example, when there is a match but the query type is _not_ for an A record, the extension will be returning a NODATA, ins ead of passing through.

Is there a reason behind this? Would it be better to define a consistent behavior across, probably gated behind. "fail_open" config knob?

extensions/dns-gateway/src/dns_gateway.rs
```diff
@@ -0,0 +1,415 @@
...
+        let config = r#"{
+            "value": {
+                "domains": [
+                    {"domain": "*", "base_ip": "10.239.0.0", "prefix_len": 24, "metadata": {}}
+                ]
+            }
+        }"#;
+
+        assert!(DnsGatewayFilterConfig::new(config.as_bytes()).is_none());
+    }
```
nacx, line 401: All the tests here are about configuration parsing, etc, but not about the plugin itself.
I'm not super versed on the Rust SDKs, but is it possible to mock this UDPListener filter to actually test what it will do on the various cases?

nacx: I see some tests are in the `lib.rs` but are quite vague and not asserting on concrete values. Is it feasible to have more precise tests?

extensions/dns-gateway/manifest.yaml
```diff
@@ -0,0 +1,122 @@
...
+  - **DNS**: UDP port 53 redirected to Envoy's DNS listener
+  - **TCP**: Outbound connections to virtual IP ranges redirected to Envoy's TCP listener
+
+  ## Building
+
+  ```bash
+  cargo build --release -p dns-gateway
+  ```
+
+  The compiled library will be at `target/release/libdns_gateway.{so,dylib}`
```
nacx: Let's remove this section. The manifest should be more about usage info. All extensions are expected to be built hte same way and that should be already covered in the BOE docs.

extensions/dns-gateway/src/dns_gateway/cache_lookup.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        let (domain, metadata) = match get_cache().lookup(ip) {
+            Some(d) => d,
+            None => {
+                envoy_log_warn!("No destination found for virtual IP: {} (cache miss)", ip);
+                return abi::envoy_dynamic_module_type_on_network_filter_data_status::Continue;
+            }
+        };
+
+        envoy_filter.set_filter_state_bytes(b"envoy.dns_gateway.domain", domain.as_bytes());
```
nacx: Can we make the prefix be `io.builtonenvoy.dns_gateway` to align with other extensions? We can make the prefix configurable to let users change it if needed.

extensions/dns-gateway/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+| Key                                | Example                          |
+| ---------------------------------- | -------------------------------- |
+| `envoy.dns_gateway.domain`         | `bucket-1.aws.com`               |
+| `envoy.dns_gateway.metadata.<key>` | value from matched domain config |
+
+Usage in Envoy config:
+
+- `%FILTER_STATE(envoy.dns_gateway.domain:PLAIN)%`
+- `%FILTER_STATE(envoy.dns_gateway.metadata.cluster:PLAIN)%`
```
nacx: I'd add this to the `manifest.yaml` so that it is rendered inht e extension doc page.

nacx (commented):
Some initial small nits

extensions/dns-gateway/manifest.yaml
```diff
@@ -0,0 +1,122 @@
+# Copyright Built On Envoy
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+name: dns-gateway
+version: 0.1.0
+minEnvoyVersion: 1.37.1
```
nacx: This should probably be bumped to `1.38.0` now?

nacx (approved):
Looks good. Thank you!

PR #368, opened 2026-04-16
feature: network filter support for CLI

ovadiagal:
Currently, only HTTP filters can be registered using the built on envoy CLI. This PR adds support for registering network filters.

This is a preflight for https://github.com/tetratelabs/built-on-envoy/pull/366, as I want to register both a network filter and a UDP listener filter. I intend to follow up with UDP listener filter support in a separate PR.


cli/cli
nacx, line 1: Can you remove this file?

nacx (commented):
This looks good to me. Thank you for this contribution!
Just one comment about a leftover, but overall LGTM.

@wbpcode mind taking a look?

cli/cli
ovadiagal: fixed

nacx:
nit: could you also include the filter type in a column in the `boe list` command output?

nacx:
And something to also add (feel free to do this in a follow-up PRs):

*  It would be great to also enhance the `boe create` command to add a `--filter-type` parameter to allow scaffolding a network filter. AFAIK it is not yet supported in the Go SDK, so we can validate that the type is Rust when using it.
* It would also be great to update the [Writing Rust Extensions](https://github.com/tetratelabs/built-on-envoy/blob/main/website/src/pages/docs/writing-rust-extensions.mdx) docs page to add a section specific to Network Filters.

wbpcode (commented):
LGTM. Thanks.

nacx (approved):

PR #373, opened 2026-04-17
cli: create network filters

AntonKanug:


cli/cmd/templates/create/rust/manifest.yaml.tmpl
```diff
@@ -3,6 +3,51 @@ version: 0.1.0
 categories:
   - Misc
 author: Unknown
+{{- if eq .FilterType "network" }}
```
nacx: Let's have also a `manifest-network.yaml.tmpl` and a `manifest.http.yaml.tmpl` for a cleaner setup

cli/cmd/templates/create/rust/lib_network.rs.tmpl
nacx, line 1: let's rename the existing one to `lib_http.rs.tmpl` for consistency?

cli/cmd/templates/create/rust/lib_network.rs.tmpl
```diff
@@ -0,0 +1,144 @@
...
+        let filter_config = FilterConfig::new("");
+        assert!(filter_config.is_some());
+        assert_eq!(filter_config.unwrap().config.log_tag, "example");
+    }
+
+    #[test]
+    fn test_new_filter_config_invalid_json() {
+        let filter_config = FilterConfig::new("{invalid");
+        assert!(filter_config.is_none());
+    }
```
nacx, line 143: AFAIK in Envoy 1.37.2` the SDK still does not provide Mocks to easily test network filters, is this correct:? (cc @wbpcode). Is there a way we can have tests for the main callbacks in this version of the SDK?

cli/cmd/create.go
```diff
@@ -40,11 +41,19 @@ func (c *Create) Help() string { return createHelp }
...
+	filterType := c.FilterType
+	if filterType == "" {
+		filterType = "http"
+	}
+
 	switch c.Type {
 	case "go":
+		if filterType == "network" {
+			return fmt.Errorf("network filter scaffolding is not supported for Go extensions; use --type rust")
+		}
```
nacx: We can probably create a `Validate` method and move the logic to that method the CLI will autoamtically call, then keep run assumign the data is already correct.

nacx (commented):
Looking good! Jus a few minor comments

cli/e2e/create_test.go
```diff
@@ -169,6 +169,59 @@ func TestCreateRustWithDockerSupport(t *testing.T) {
 	})
 }
 
+func TestCreateRustNetworkWithDockerSupport(t *testing.T) {
```
nacx, line 172: I would remove this test. It does not test anything additional to the existing one, where we test the pushed manifest, image, etc. I think with the unit tests we have in place the feature is already properly covered.

nacx: Actually, let's keep it as it will exercise the extension `build`.

cli/cmd/templates/create/rust/lib_network.rs.tmpl
AntonKanug: yup, there is no mocks for EnvoyNetworkFilter

>  Is there a way we can have tests for the main callbacks in this version of the SDK?

yes but you have to hand roll the mocks, should i add that?

AntonKanug: https://github.com/envoyproxy/envoy/pull/44514

nacx: Let's add this when it is released in Envoy. Thanks for doing the upstream PR!

nacx (approved):

PR #387, opened 2026-04-24
feat(composer): update Envoy version requirements to 1.38.0 and bump dynamic modules dependency

sjoukedv:
Make the latest envoy version available: https://github.com/envoyproxy/envoy/releases/tag/v1.38.0

### DIFF
--- a/extensions/composer/anthropic-decoder/plugin_test.go
+++ b/extensions/composer/anthropic-decoder/plugin_test.go
@@ -47,10 +47,17 @@ func defaultCfg() *anthropicDecoderConfig {
 	return &anthropicDecoderConfig{MetadataNamespace: defaultMetadataNamespace}
 }
 
+func newMockHttpFilterHandle(ctrl *gomock.Controller) *mocks.MockHttpFilterHandle {
+	h := mocks.NewMockHttpFilterHandle(ctrl)
+	h.EXPECT().ReceivedBufferedRequestBody().Return(true).AnyTimes()
+	h.EXPECT().ReceivedBufferedResponseBody().Return(true).AnyTimes()
+	return h
+}
+
 // newPluginHandleWithoutPerRouteConfig creates a mock HttpFilterHandle with default expectations including
 // GetMostSpecificConfig returning nil (no per-route config).
 func newPluginHandleWithoutPerRouteConfig(ctrl *gomock.Controller) *mocks.MockHttpFilterHandle {
-	pluginHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	pluginHandle := newMockHttpFilterHandle(ctrl)
 	pluginHandle.EXPECT().Log(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	pluginHandle.EXPECT().GetMostSpecificConfig().Return(nil).AnyTimes()
 	return pluginHandle
@@ -59,7 +66,7 @@ func newPluginHandleWithoutPerRouteConfig(ctrl *gomock.Controller) *mocks.MockHt
 // newPluginHandleWithPerRouteConfig creates a mock HttpFilterHandle that returns
 // the given per-route config from GetMostSpecificConfig.
 func newPluginHandleWithPerRouteConfig(ctrl *gomock.Controller, perRouteConfig any) *mocks.MockHttpFilterHandle {
-	pluginHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	pluginHandle := newMockHttpFilterHandle(ctrl)
 	pluginHandle.EXPECT().Log(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	pluginHandle.EXPECT().GetMostSpecificConfig().Return(perRouteConfig).AnyTimes()
 	return pluginHandle
@@ -220,7 +227,7 @@ func TestWellKnownHttpFilterConfigFactories(t *testing.T) {
 func TestOnRequestHeaders_EndOfStream(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	mockHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	mockHandle := newMockHttpFilterHandle(ctrl)
 
 	filter := &decoderFilter{handle: mockHandle, config: defaultCfg()}
 	result := filter.OnRequestHeaders(fake.NewFakeHeaderMap(map[string][]string{}), true)
@@ -230,7 +237,7 @@ func TestOnRequestHeaders_EndOfStream(t *testing.T) {
 func TestOnRequestHeaders_NotEndOfStream(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	mockHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	mockHandle := newMockHttpFilterHandle(ctrl)
 
 	filter := &decoderFilter{handle: mockHandle, config: defaultCfg()}
 	result := filter.OnRequestHeaders(fake.NewFakeHeaderMap(map[string][]string{}), false)
@@ -242,7 +249,7 @@ func TestOnRequestHeaders_NotEndOfStream(t *testing.T) {
 func TestOnRequestBody_NotEndOfStream(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	mockHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	mockHandle := newMockHttpFilterHandle(ctrl)
 
 	filter := &decoderFilter{handle: mockHandle, config: defaultCfg()}
 	result := filter.OnRequestBody(newTestBodyBuffer([]byte("data")), false)
@@ -252,7 +259,7 @@ func TestOnRequestBody_NotEndOfStream(t *testing.T) {
 func TestOnRequestBody_EmptyBody(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	mockHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	mockHandle := newMockHttpFilterHandle(ctrl)
 
 	mockHandle.EXPECT().BufferedRequestBody().Return(newTestBodyBuffer([]byte{})).AnyTimes()
 	mockHandle.EXPECT().ReceivedRequestBody().Return(nil).AnyTimes()
@@ -265,7 +272,7 @@ func TestOnRequestBody_EmptyBody(t *testing.T) {
 func TestOnRequestBody_InvalidJSON(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	mockHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	mockHandle := newMockHttpFilterHandle(ctrl)
 
 	body := []byte(`{invalid json}`)
 	mockHandle.EXPECT().BufferedRequestBody().Return(newTestBodyBuffer(body)).AnyTimes()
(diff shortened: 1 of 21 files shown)


sjoukedv:
@nacx should we also release this as `0.6.0`?

sjoukedv:
Once `github.com/envoyproxy/go-control-plane/envoy` is updated we should also update it here

extensions/composer/go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/corazawaf/coraza/v3 v3.7.0
 	github.com/crewjam/saml v0.5.1
 	github.com/docker/cli v29.4.0+incompatible
-	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260311012303-5ef4e4cea57f
+	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260423231439-f1dd21b16c24
```
nacx, line 17: Can you also update it in the different create command templates?
That should fix the e2e failure, I think, as they scaffold a new extension and that would be using the older SDK, meaning hte go plugin is built with a different SDK than the composer, causing the e2e failure.

nacx:
> Once github.com/envoyproxy/go-control-plane/envoy is updated, we should also update it here

Yes! And once that is available, there are also some small improvements, like being able to prefix the stats with a proper prefix that we'll be able to add.

sjoukedv:
> > Once github.com/envoyproxy/go-control-plane/envoy is updated, we should also update it here
> 
> Yes! And once that is available, there are also some small improvements, like being able to prefix the stats with a proper prefix that we'll be able to add.

I'll leave this open and keep an eye out for the new version. I have already tested the composer with 1.38 which works like a charm.

extensions/composer/go.mod
sjoukedv: Yup, that was it. It's fixed now with all tests passing.

sjoukedv:
> Once `github.com/envoyproxy/go-control-plane/envoy` is updated we should also update it here

@nacx shall we merge? dependatbot can detect this dependency.


Does this also replace https://github.com/tetratelabs/built-on-envoy/pull/354?

nacx (commented):
Can you uadpte this as well? https://github.com/tetratelabs/built-on-envoy/blob/main/cli/cmd/templates/create/rust/Cargo.toml.tmpl#L10

sjoukedv:
> Can you uadpte this as well? https://github.com/tetratelabs/built-on-envoy/blob/main/cli/cmd/templates/create/rust/Cargo.toml.tmpl#L10

Done

extensions/composer/anthropic-decoder/plugin_test.go
```diff
@@ -47,10 +47,17 @@ func defaultCfg() *anthropicDecoderConfig {
 	return &anthropicDecoderConfig{MetadataNamespace: defaultMetadataNamespace}
 }
 
+func newMockHTTPFilterHandle(ctrl *gomock.Controller) *mocks.MockHttpFilterHandle {
+	h := mocks.NewMockHttpFilterHandle(ctrl)
+	h.EXPECT().ReceivedBufferedRequestBody().Return(true).AnyTimes()
+	h.EXPECT().ReceivedBufferedResponseBody().Return(true).AnyTimes()
+	return h
+}
```
nacx, line 52: Instead of doing this, can we add the explicit expectation to the tests? The main difference is that this in practice _ignores_ this check, but having it explicitly on each test makes it aprt of the test contract and helps catch regressions.

extensions/composer/chat-completions-decoder/plugin_test.go
```diff
@@ -47,20 +47,31 @@ func defaultCfg() *chatCompletionsDecoderConfig {
...
 	return pluginHandle
 }
 
 // newPluginHandleWithPerRouteConfig creates a mock HttpFilterHandle that returns
 // the given per-route config from GetMostSpecificConfig.
 func newPluginHandleWithPerRouteConfig(ctrl *gomock.Controller, perRouteConfig any) *mocks.MockHttpFilterHandle {
-	pluginHandle := mocks.NewMockHttpFilterHandle(ctrl)
+	pluginHandle := newMockHTTPFilterHandle(ctrl)
 	pluginHandle.EXPECT().Log(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+	pluginHandle.EXPECT().ReceivedBufferedRequestBody().Return(true).AnyTimes()
```
nacx: Here and in the other places where this is set individually on each test, do we really need the `AnyTimes()`?

nacx (commented):
Just some final comments

nacx:
>Does this also replace https://github.com/tetratelabs/built-on-envoy/pull/354?

I don't think it replaces it, as in that PR there are additional changes like removing the custom test buffers and using hte ones from the SDK.

extensions/composer/chat-completions-decoder/plugin_test.go
sjoukedv: Nope, changed

extensions/composer/anthropic-decoder/plugin_test.go
sjoukedv: Done

nacx (dismissed):

extensions/composer/manifest.yaml
```diff
@@ -6,7 +6,7 @@
 name: composer
 version: 0.6.0-dev
 composerVersion: 0.6.0-dev
-minEnvoyVersion: 1.37.1
+minEnvoyVersion: 1.38.0
```
nacx, line 9: With the bump of the SDK, we are mostly changing tests, but there was the idea that the SDK would remain compatible starting 1.37.1.

Do you know if Envoy 1.38 is needed? Or given the SDK is compatible now, could we keep this as 1.37.1?
Could you try running it with Envoy 1.37.1 and see if there are issues?

nacx (commented):
Sorry, I just caught this and wanted us to have a good understanding.

No issues with the PR, it's good to go, but I want to understand the real min Envoy version before merging.

extensions/composer/manifest.yaml
sjoukedv: You're right the docs say it here https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.37/v1.37.1#new-features. CI was failing before with 1.37.0 which did not include the compatibility yet. If this works with 1.37.1 (or .2) we can also re-include this in the test matrix envoy versions.

sjoukedv: Okay, so it was hard to wrap my head around, but the minVersion of `1.38.0` is correct as there is backwards compatibility. We are now bumping the composer dynamic module to that of `1.38.0` which is also compatible with `1.39.0`. This is the reason why the integration tests run against the latest stable version + `-dev`  and it keeps working.

It would be possible to keep the existing dynamic module version (that of `1.37.1`) and only bump envoy to `1.38.0`, but than we can't use the new features introduced. Because the dynamic module version is directly coupled to the envoy version it makes sense to me to update them both. 

Since it's possible to use e.g. `0.6.0-dev` with `1.38.0`, but not with `1.39.0`, we can make the envoy compatibility matrix more clear instead of just `>= 1.37.1` it should be `>= 1.37.1 && <= 1.38.x`. I will create an issue for this.

nacx: As far as I see in this PR, most of the changes to the code are in the test packages, so probably we can change the composer manifest to have the following?
```
minEnvoyVersion: 1.37.1
maxEnvoyVersion: 1.38.0
```
We don't know if future 1.38 versions may break it, so I think this is a reasonable default.

Also, if some composer plugins use specific new features, they can individually override these properties and those should be respected by BOE.

WDYT?

sjoukedv: I like the max version, hence also the issue raised, but since this PR also bumps the dynamic module version it should be min 1.38.0 and max 1.39.x. as it's not compatible with 1.37.x anymore because of new functionality introduced.

Can we resolve that as part of https://github.com/tetratelabs/built-on-envoy/issues/389 ?

sjoukedv: The max version should also be calculated automatically ideally by using the current minor+1

nacx: >Can we resolve that as part of https://github.com/tetratelabs/built-on-envoy/issues/389 ?

Yes.

nacx: > this PR also bumps the dynamic module version it should be min 1.38.0

But my question is, even with the dynamic module version bump, my understanding was that 1.38 DYM was compatible with 1.37.1. We can resolve the versioning as part of hte linked issue and relax the constraints once we can be more granular about versions, but just want to understand if the composer plugins can still run in 1.37.1 after this is merged.

sjoukedv: My understanding is that it can't run with 1.37.1 anymore. I also couldn't get it working (please let me know if you can). It's forward compatible not backwards so the version we had was already working with 1.38.x (and is tested in CI with distroless-dev), but because it also bumps the dynamic modules library to that of 1.38.x the compatible versions become 1.38.x and 1.39.x. Technically we could have left the dynamic module library the old one and only bump envoy version itself. That would mean we can't develop against newer dynamic modules functionality that were only introduced in 1.38.0.

wbpcode: The compatibility works in following way:

1. if you module is compiled with 1.38's SDK, it should still could be loaded and executed in the all newer Envoy version like 1.39, 1.40 and so on.
2. If you upgraded your SDK to the 1.39's SDK, then the module may cannot be loaded and executed by 1.38 because the SDK may have used the latest ABI that introduced in the Envoy 1.39.

I haven't figured out a perfect solution to resolve the conflicts between the latest feature support and more widely Envoy versions support except to create an independent release branch for every maintained Envoy version, Esp for the golang because all golang extensions need to be compiled into single binary.

wbpcode: > but just want to understand if the composer plugins can still run in 1.37.1 after this is merged.

Nope. So, I will suggest always keep to use the oldest SDK version that still be maintained and could meet our requirements. Don't easily to upgrade the SDK version except:

1. CVE patches.
2. explicit new feature requirement.
3. the existing SDK is out of support.

Or we could create specific branch for every Envoy version, or we could create subdirectory of `composer-1-37`, `composer-1-38` and so on to maintain them at same branch.

nacx: That makes it very clear. Thanks!!

Also good to know about the considerations to upgrade the SDK. We should add it to a markdown file in the root of the repo with some explanation.

nacx (approved):

extensions/composer/manifest.yaml
sjoukedv: > > but just want to understand if the composer plugins can still run in 1.37.1 after this is merged.
> 
> Nope. So, I will suggest always keep to use the oldest SDK version that still be maintained and could meet our requirements. Don't easily to upgrade the SDK version except:
> 
> 1. CVE patches.
> 2. explicit new feature requirement.
> 3. the existing SDK is out of support.
> 
> Or we could create specific branch for every Envoy version, or we could create subdirectory of `composer-1-37`, `composer-1-38` and so on to maintain them at same branch.

I would argue for upgrading the SDK to latest envoy version that is released, not only for the exceptions you mention above. Main point would be that you can still use the existing versions of extensions with the older envoy version as they were built against the older SDK. Newer release versions of extensions are built against the latest SDK, so if you want to use them you need to upgrade envoy to that version (everbody hopefully upgrades their envoy at some point anyway because of CVEs and/or bug fixes). We can provide documentation on how to build against older envoy versions or backport any changes to x-1 (which is more effort).

wbpcode: I think we need to evaluate the gain of the upgrading. Because in most cases, the upgrading brings nothing but will narrow the compatibility.

I didn't see high priority feature requirement that cannot addressed by the current SDK. Upgrading to 1.38 may be fine because we marked 1.38 as GA for HTTP DYM. But we still need to note, in most cases, for most users, they won't use the latest Envoy immediately. And in most cases, they may have multiple versions at same time for a long time. How to keep them get same extensions list for multiple versions will be a problem.

Esp for composer/golang, because it's an extension bundle.

PR #396, opened 2026-04-28
[bp/composer/0.5.0] [stack] back port necessary fix of WAF to composer v0.5.0

wbpcode:


nacx (approved):

nacx:
Many workflows filter branches with prefix `release-**`. Should we update those prefixes to be `release/**` instead? This can be done in another PR.

wbpcode:
> Many workflows filter branches with prefix `release-**`. Should we update those prefixes to be `release/**` instead? This can be done in another PR.

I was working on that but then stopped because a discussion about CVE from Envoy community. I will hold this PR and until our workflow could works to release branch first.

PR #397, opened 2026-04-28
extension: allow manifests to declare native HTTP filters before themselves

codefromthecrypt:
Add nativeHttpFilters.before to the manifest schema and validation. Extensions that produce an HTTP filter can list native Envoy filters to emit ahead of their own in the HCM chain. Validation rejects reserved names, duplicates, and the field on extension types with no HTTP anchor.

Generate blank imports of all go-control-plane HTTP filter proto packages so any `@type` in a native filter `typed_config` resolves at runtime.

Add `--native-http-filter-before` to boe run and boe gen-config as a repeatable runtime override (inline YAML/JSON or `@filepath`).

Embed manifest.yaml in OCI artifact images so `nativeHttpFilters` data travels with the artifact. Merge OCI annotation metadata into the embedded manifest on pull when Version or ComposerVersion are empty


cli/internal/envoy/config.go
```diff
@@ -20,6 +20,10 @@ import (
 	listenerv3 "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	routev3 "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	streamv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/access_loggers/stream/v3"
+	dymhttpv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/dynamic_modules/v3"
+	// Imported for protojson @type registration so manifests can declare
+	// `envoy.filters.http.mcp` in nativeHttpFilters.before.
+	_ "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/mcp/v3"
```
nacx: Should we have some codegen similar to this: https://github.com/envoyproxy/gateway/blob/6774309541283e4fef49f452682d3adc6e41b7bc/internal/xds/extensions/extensions.go
To automatically generate all these imports and have all filters available?

extensions/composer/Dockerfile.plugin
```diff
@@ -24,4 +24,11 @@ RUN CGO_ENABLED=1 go build -trimpath -buildmode=plugin ${BUILD_TAG} -o plugin.so
...
 FROM scratch AS final
 
+ARG EXTENSION_PATH
+
 COPY --from=builder /workspace/plugin.so /
+# Preserve the plugin's manifest.yaml so downloaded artifacts retain
+# manifest-owned fields (e.g. nativeHttpFilters) that the OCI annotation
+# fallback cannot represent. /workspace/manifest.yaml is the composer
+# uber-manifest; the plugin manifest lives under $EXTENSION_PATH.
+COPY --from=builder /workspace/${EXTENSION_PATH}/manifest.yaml /
```
nacx, line 28: The composer-based extension manifests may have incomplete information in the source tree, as they inherit things like the version from the parent composer manifest.

When the downloader downloads an extension, if the package contains a manifest, it loads it and replaces the info from the OCI annotations. 

https://github.com/tetratelabs/built-on-envoy/blob/d43ae72fb941585adf4f05e1fdbd4de9afd108c2/cli/internal/extensions/downloader.go#L91-L100

This could be problematic with composer extensions that package the manifests without the `version` and `composerVersion` properties set.

We could do a couple of things:

* Update that code to preserve the `version` and `composerVersion` from the OCI annotations, as it should always be correct there.
* Trigger somehow proper manifest generation (so that the manifest that is bundled in the package contains the version, composerVersion, and min/max Envoty versions). This would be ideal (and min/max envoy versions are already missing as they're not in the OCI annotations), but would require more plumbing.

I think either approach is fine for this PR.

nacx (commented):
Thanks!

I've left a couple of comments.

Apart from that, more on a design angle, do you see the native filters as something that would go on the extension manifest? Or given that they contain configuration that could be different between runs, it would belong more to a "runtime config" that should be fed to the `run` command on the fly?

codefromthecrypt:
thanks for the review @nacx  I had left it draft because indeed I needed direction, just you beat me to asking for it. Basically, I currently don't care outside mcp and this works for code I'm playing with locally. Anything beyond that is probably your call so I'll defer to your advice and you mention the direction that makes most sense.

codefromthecrypt:
@nacx here are some things. I ran out of time to manually integration test but hope this is in better shape. I updated the pr desc.

## Runtime config flag

Added `--native-http-filter-before` to boe run and boe gen-config. Repeatable per-extension position, accepts YAML/JSON or @filepath. This overrides the manifest's nativeHttpFilters.before when provided; otherwise the manifest default is used.

## Codegen for proto imports.

Replaced the manual MCP blank import with a `go:generate` directive that produces blank imports of all go-control-plane HTTP filter packages.

## Downloader merge.

After loading the embedded manifest from a pulled artifact, Version and ComposerVersion from the OCI annotations are preserved when the embedded YAML has them empty.

Also embedded manifest.yaml in OCI artifact Dockerfiles (plugin, rust, rust template) so nativeHttpFilters data travels with the artifact without annotation encoding. Integration test covers the round-trip.

codefromthecrypt:
doing one last check against my local POC and will pull out of draft if all good

nacx (approved):
Thanks!

PR #398, opened 2026-04-28
make workflows work for release/** branches

wbpcode:


.github/workflows/release-extensions.yaml
```diff
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - release/**
```
nacx, line 12: This will only trigger those extensions where there are changes in the commit right?
I think this is the case, but just want to confirm that having this won't open the door to some extension push to the registry from the local branch, to a "main" extension.

nacx (approved):
LGTM. Just one comment.

.github/workflows/release-extensions.yaml
wbpcode: Yeah. This only release extensions that versions have changed in the latest commit. And a fresh new branch will be trigger this workflow because the push.paths filter, until there is new commit be pushed into the release branch.

PR #402, opened 2026-05-03
docs: add implicit maxEnvoyVersion calculation and update display logic

### LINKED ISSUE #389: More strict dynamic module envoy version constraints
Following from: https://github.com/tetratelabs/built-on-envoy/pull/387#pullrequestreview-4172434839

It's possible to use e.g. 0.6.0-dev with 1.38.0, but not with 1.39.0, we can make the envoy compatibility matrix more clear instead of just >= 1.37.1 it should be >= 1.37.1 && <= 1.38.x.

It would be great to also add versioning to the docs website so we can view previous versions. Right now, it's always showing `-dev` instead of latest "stable" version.

sjoukedv:
Fixes https://github.com/tetratelabs/built-on-envoy/issues/389

image

The question mark links to https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules#compatibility

Co-authored-by: Copilot <copilot@github.com> Signed-off-by: Sjouke de Vries <info@sdvservices.nl>


website/src/pages/extensions/[name].astro
```diff
@@ -38,6 +38,14 @@ export async function getStaticPaths() {
 
 const { extension } = Astro.props as { extension: Extension };
 
+// When only minEnvoyVersion is set, compute an implicit max as major.(minor+1).patch
+function implicitMaxEnvoyVersion(minVersion: string): string {
```
Comment (nacx) [inline website/src/pages/extensions/[name].astro]: Instead of doing this only for the catalog display, can we create the `implicitMaxEnvoyVersion` function in the [cli/extensions/manifest.go](https://github.com/tetratelabs/built-on-envoy/blob/main/cli/internal/extensions/manifest.go) file, and call it (logic should change to use `.0` instead of `.x` as the patch version):

Here, if the `minEnvoyVersion` is set and `maxEnvoyVersion` is blank:
https://github.com/tetratelabs/built-on-envoy/blob/5d8947fdddf0e93a8223101b0effb5586dcd3c0d/cli/internal/extensions/manifest.go#L143-L148

And here, just after resolving the versions from the composer parent, (if the `minEnvoyVersion` is set and `maxEnvoyVersion` is blank):
https://github.com/tetratelabs/built-on-envoy/blob/5d8947fdddf0e93a8223101b0effb5586dcd3c0d/cli/internal/extensions/manifest.go#L287-L308

By having the function here and calling it in these two places, the "runtime manifests" should always have the max version properly set, and that should also be automatically populated to the generated catalog.

We shouldn't need custom display logic, and all code would be consistent.

nacx (commented):
Thanks! Added one comment

website/src/pages/extensions/[name].astro
Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro]: I explicitly set it to `major.minor.x` for the upper limit since strictly `<= major.minor.0` does not include the patch versions that are later on introduced.

The relevant logic has been moved to the CLI instead of the website folder.

Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro]: Do you prefer `.x` or `.0` for the upper constraint?

Comment (nacx) [reply website/src/pages/extensions/[name].astro]: I'd say we should use `.0` as that is a real release.

We can also update the semantics for the Envoy version interval to be `[min, max)` so that the max version is not included. WDYT?

cli/internal/extensions/manifest.go
```diff
@@ -140,11 +141,28 @@ func (m *Manifest) EnvoyConstraints() string {
 	return constraints
 }
 
+// ImplicitMaxEnvoyVersion computes an implicit max Envoy version from a min version.
+// Given a semver like "1.33.0", it returns "1.34.0" (major.(minor+1).0).
+func ImplicitMaxEnvoyVersion(minVersion string) string {
+	parts := strings.SplitN(minVersion, ".", 3)
+	if len(parts) < 2 {
+		return minVersion
```
nacx: If we can't parse the min version, we should probably leave the max version empty, as we can't effectively compute it to a good value.

cli/internal/extensions/manifest.go
```diff
@@ -140,11 +141,28 @@ func (m *Manifest) EnvoyConstraints() string {
...
+	}
+	minor, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return minVersion
+	}
+	return fmt.Sprintf("%s.%d.x", parts[0], minor+1)
```
nacx: Let's use `.0` as it is a real version?
```suggestion
	return fmt.Sprintf("%s.%d.0", parts[0], minor+1)
```

website/src/pages/extensions/[name].astro
Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro]: The problem I keep having with `.0` is that mathematically the constraint it not valid anymore with `<=` .

You're right that `.x` is not a "real" version, but I have seen letters being used more often as a placeholder in https://semver.org/.

cli/internal/extensions/manifest.go
sjoukedv: If we accept this change the changes to the manifest schema can also be reverted.

website/src/pages/extensions/[name].astro
Comment (nacx) [reply website/src/pages/extensions/[name].astro]: >We can also update the semantics for the Envoy version interval to be [min, max) so that the max version is not included. WDYT?

Yeah, this is why I htink it is OK to change the semantics and assume min version is included, but max version is not. WDYT?

cli/internal/extensions/manifest.go
nacx: SGTM

website/src/pages/extensions/[name].astro
Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro]: Can you provide the implementation of this?

Comment (nacx) [reply website/src/pages/extensions/[name].astro]: Sure! I can do in a follow-up PR

nacx (commented):
PR looks good.
A couple of final minors:

* Can you add a link to the Envoy docs around version compatibility in the `ImplicitMaxEnvoyVersion` doc?
* Can you make the function return `""` when the version does not match the format? It should never happen as the values are validated, but it's a more reasonable value to return on error, I think.
* Can you add unit tests for the `ImplicitMaxEnvoyVersion` function?

sjoukedv:
> PR looks good. A couple of final minors:
> 
> * Can you add a link to the Envoy docs around version compatibility in the `ImplicitMaxEnvoyVersion` doc?
> * Can you make the function return `""` when the version does not match the format? It should never happen as the values are validated, but it's a more reasonable value to return on error, I think.
> * Can you add unit tests for the `ImplicitMaxEnvoyVersion` function?

Done, let me know if I forgot something

nacx (approved):
Thanks!  LGTM.
Looks like the PR needs to be rebased on top of latest main

sjoukedv:
My bad, replaced by https://github.com/tetratelabs/built-on-envoy/pull/414

PR #407, opened 2026-05-06
Add docker HEALTHCHECK

codefromthecrypt:
This adds a `boe healthcheck` subcommand basically the same way I wrote it for envoy AI gateway.

The subcommand uses func-e's admin client to find Envoy under pid 1 and returns success when Envoy admin readiness succeeds.

I also added `ENVOY_LOG_LEVEL` to `boe run` to reduce plumbing in external projects.


cli/Dockerfile
```diff
@@ -57,4 +57,8 @@ COPY --chown=boe:boe ./out/boe-${TARGETOS}-${TARGETARCH} /boe
 COPY --chown=boe:boe ./entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
+# The healthcheck subcommand discovers Envoy's admin server and checks readiness.
+HEALTHCHECK --interval=10s --timeout=5s --start-period=5s --retries=3 \
+    CMD ["/boe", "healthcheck"]
```
nacx, line 62: Hmmmm, can you elaborate on the main use case for the command? I'm trying to understand what this is trying to provide, as I think it will not work as expected.

`boe run` takes care of starting Envoy, etc. Envoy is not running _before_ the `boe run` command is executed, and its lifecycle is tied to that command.

IIUC the code, when the healthcheck command is executed at this point, there will be no Envoy process running so it will always fail?

cli/cmd/healthcheck.go
```diff
@@ -0,0 +1,47 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
+	defer cancel()
+
+	logger := slog.New(slog.NewTextHandler(stderr, &slog.HandlerOptions{}))
+	// In docker, pid 1 is the boe process
+	return doHealthcheck(ctx, 1, logger)
+}
+
+func doHealthcheck(ctx context.Context, boePid int, logger *slog.Logger) error {
+	if adminClient, err := admin.NewAdminClient(ctx, boePid); err != nil {
```
codefromthecrypt, line 39: @nacx this is the tricky bit. under the assumption that boe is pid 1, its direct and only child is envoy. when the healthcheck command runs, it is a subprocess, but traces back not from its pid, but pid 1.

So, that's why it works here (and aigw where this approach was lifted from)

cli/Dockerfile
```diff
@@ -57,4 +57,8 @@ COPY --chown=boe:boe ./out/boe-${TARGETOS}-${TARGETARCH} /boe
...
+
 ENTRYPOINT ["/entrypoint.sh"]
```
codefromthecrypt, line 64: the main difference here is aigw defaults to "run" and so the HEALTHCHECK makes more sense when that's the case I guess https://github.com/envoyproxy/ai-gateway/blob/main/Dockerfile#L59

nacx: This image exists mostly because there are no Envoy binaries for `darwin/amd64`, so users of that platform can do `boe run --docker` or `boe gen-config --docker` (IIRC there are no other use cases beyond these two).
For the first command, it should work fine, but for hte `gen-config` the healthcheck may fail. I don't think it's a big deal, as we don't care much about the health status of this container when using hte `gen-config` command, but I'm not knowledgeable enough about how these health checks work. Would that have any side-effect?

cli/cmd/root.go
```diff
@@ -30,13 +30,14 @@ const CLIName = "boe"
...
-	List      List      `cmd:"" help:"List available extensions"`
-	Run       Run       `cmd:"" help:"Run Envoy with extensions"`
-	GenConfig GenConfig `cmd:"" help:"Generate Envoy configuration with extensions"`
-	Create    Create    `cmd:"" help:"Create a new extension template"`
-	Download  Download  `cmd:"" help:"Download extensions from the registry"`
-	Clean     Clean     `cmd:"" help:"Clean cache directories"`
-	Version   Version   `cmd:"" help:"Print version information"`
+	List        List        `cmd:"" help:"List available extensions"`
+	Run         Run         `cmd:"" help:"Run Envoy with extensions"`
+	Healthcheck Healthcheck `cmd:"" help:"Docker HEALTHCHECK command."`
```
nacx: I guess this command is only used internally by the Docker image and is not meant for users to use, as it hardcodes PID 1.

Does it make sense to add `hidden:""` so it is skipped from the help output?
If we add this, then we should also change this condition:
https://github.com/tetratelabs/built-on-envoy/blob/0250ec349261691f3478ff2d4be6c386c8d1c4e2/cli/tools/gen-cli-docs/main.go#L198
To something like:
```go
if node.Type != kong.CommandNode || node.Hidden {
```
So that the command is not included in hte docs site.

WDYT?

cli/cmd/healthcheck.go
nacx: Oh, got it. Thanks for hte explanation!

cli/cmd/root.go
codefromthecrypt: sure!

codefromthecrypt:
when done, I'll make sure aigw has the same hidden thing. I agree.

nacx (approved):
Nice, thanks!

Issue #355, opened 2026-04-10
go dynamic module first to improve startup experience


wbpcode:
For now, the new out-of-tree extensions will be compiled as go plugin and then load it by the goplugin-loader implicitly. This design brings two problems:

1. potential dependency conflict of go plugin mechanism.
2. only one golang dynamic module extension could be loaded.

These problems are inevitable because the golang runtime/go plugin 's limits. But users/developers needn't to face these problems initially. In most cases, users/developers develop, run, test single extension at one time, it's unnecessary to care the runtime conflict. Compiling the local extension into normal golang dynamic module should be good enough to get fluent startup experience.

If they do want to deploy their custom extensions, then we can expose these problems (Error message in Envoy, validation of warning in control plane, Tips when running the BoE cli)

Thanks great feedback from @codefromthecrypt

wbpcode:
cc @nacx

nacx:
Agree.

Still, I think there is value in keeping the current code structure with embedded/standalone so that it is easier to contribute back to BOE, and have the extension ready to work in both modes. WDYT?

wbpcode:
Sure, @nacx

PR #409, opened 2026-05-06
make cshared mode be preferred for golang extension

wbpcode:
- prefer to compile local extension as shared lib rather then goplugin to avoid dependency problem in local development phase. see https://github.com/tetratelabs/built-on-envoy/issues/355


cli/cmd/templates/create/go/main_cshared.go.tmpl
```diff
@@ -0,0 +1,16 @@
...
+package main
+
+import (
+	_ "github.com/envoyproxy/envoy/source/extensions/dynamic_modules/sdk/go/abi"
+	sdk "github.com/envoyproxy/envoy/source/extensions/dynamic_modules/sdk/go"
+	impl "{{ .Name }}"
+)
+
+func init() {
+	sdk.RegisterHttpFilterConfigFactories(impl.WellKnownHttpFilterConfigFactories())
```
nacx: nit: instead of doing this, what about importing the `embedded` package that is also generated? This would lead to a `main` that is more similar to the ones people would create to import composer extensions.

wbpcode:
gently ping for a review @nacx 🙇

cli/cmd/templates/create/go/main_cshared.go.tmpl
```diff
@@ -0,0 +1,16 @@
+// Package main builds a c-shared library that can be loaded directly by Envoy as a dynamic module.
+// Use this entry point when building the extension as an independent shared library.
+// For building as a Go plugin loaded by composer, use the standalone/ directory instead.
+package main
+
+import (
+	_ "github.com/envoyproxy/envoy/source/extensions/dynamic_modules/sdk/go/abi"
+	sdk "github.com/envoyproxy/envoy/source/extensions/dynamic_modules/sdk/go"
```
nacx: nit: worth adding here a placeholder comment with some examples of importing other composer extensions?
This way users would know how to produce their custom DYM, including their extension and other stuff.
We could also add a link in the comment to this apge/section: https://builtonenvoy.io/docs/packaging-prod/#appendix-building-your-own-dynamic-module-with-a-subset-of-extensions 
WDYT?

nacx (approved):
Nice!

PR #410, opened 2026-05-06
upgrade composer's SDK to support more ABI exposure

wbpcode:
- upgraded the SDK to support get buffer limit/set buffer limit and so on features

### DIFF
--- a/cli/cmd/templates/create/go/go.mod.tmpl
+++ b/cli/cmd/templates/create/go/go.mod.tmpl
@@ -3,6 +3,6 @@ module {{ .Name }}
 go 1.26.2
 
 require (
-	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260423231439-f1dd21b16c24
+	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260425100846-90594f45b3ed
 	github.com/stretchr/testify v1.11.1
 )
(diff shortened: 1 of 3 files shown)


wbpcode:
cc @M4tteoP here. Once this done, we can fix WAF's buffer problem easily :)

cli/cmd/templates/create/go/go.mod.tmpl
```diff
@@ -3,6 +3,6 @@ module {{ .Name }}
 go 1.26.2
 
 require (
-	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260423231439-f1dd21b16c24
+	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260425100846-90594f45b3ed
```
nacx, line 6: Can you also udpate the Rust tempalte?
https://github.com/tetratelabs/built-on-envoy/blob/0250ec349261691f3478ff2d4be6c386c8d1c4e2/cli/cmd/templates/create/rust/Cargo.toml.tmpl#L10

wbpcode: At commit f1dd21b16c244bda00edfb5ffce577e12d0d2ec2 (1.38.0), the rust SDK has supported all features that ABI provided. But the go SDK doesn't.
So, I guess we only need to upgrade go sdk to `90594f45b3ed` (1.39.0-dev but be compatible with Envoy 1.38)?

nacx: Got it. Thanks for the explanation! :D

nacx (approved):

PR #414, opened 2026-05-08
docs: add implicit max Envoy version calculation and update extension metadata

sjoukedv:
Replaces https://github.com/tetratelabs/built-on-envoy/pull/402

Signed-off-by: Sjouke de Vries <info@sdvservices.nl>


website/src/pages/extensions/[name].astro
```diff
@@ -94,7 +94,13 @@ function formatTypeName(type: string): string {
 										<span class="meta-value">
 											{extension.minEnvoyVersion && `>= ${extension.minEnvoyVersion}`}
 											{extension.minEnvoyVersion && extension.maxEnvoyVersion && <br/>}
-											{extension.maxEnvoyVersion && `<= ${extension.maxEnvoyVersion}`}
+											{extension.minEnvoyVersion && extension.maxEnvoyVersion && (
```
Comment (nacx) [inline website/src/pages/extensions/[name].astro]: Having a closer look at this, if an extension _only_ has a `maxEnvoyVersion`, then the version won't be shown.
I know that this is a case we should probably never have, but the code should cover that. Could you update it so that the maxVersion is always shown if present, then add the help link whenever it is appropriate?

Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro]: I'll make the condition `OR` then, because max is implicitly calculated from min, and if max is only defined we can still show the forward compatible note.

Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro]: There is also the manifest schema that enforces this version; https://github.com/tetratelabs/built-on-envoy/blob/main/extensions/manifest.schema.json#L31-40, so that an extension needs to have a min version? Neverthless, the logic can be both explicit with `&&` or with `||` I think.

website/src/pages/extensions/[name].astro
```diff
@@ -94,7 +94,13 @@ function formatTypeName(type: string): string {
...
 											{extension.minEnvoyVersion && `>= ${extension.minEnvoyVersion}`}
 											{extension.minEnvoyVersion && extension.maxEnvoyVersion && <br/>}
-											{extension.maxEnvoyVersion && `<= ${extension.maxEnvoyVersion}`}
+											{(extension.minEnvoyVersion || extension.maxEnvoyVersion) && (
+												<Fragment>
+													{`<= ${extension.maxEnvoyVersion}`}
+													{' '}
+													<a href="https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules#compatibility" target="_blank" rel="noopener noreferrer" title="Envoy compatibility" style="font-size: 0.75em; vertical-align: super; text-decoration: none;">?</a>
+												</Fragment>
+											)}
```
Comment (nacx) [inline website/src/pages/extensions/[name].astro:100]: I think you still need to include the "individual" printing of the max version. Otherwise, if an extension only defines a `minEnvoyVersion`, you'd be printing still the `maxVersion` and that would be incorrect.

Given how we generate the `extensions.json` today, that can't happen, but we shouldn't have the rendering coupled to that internal detail.

Something like this should work in all cases:


```suggestion
											{extension.maxEnvoyVersion && `<= ${extension.maxEnvoyVersion}`}
											{(extension.minEnvoyVersion || extension.maxEnvoyVersion) && (
												<a href="https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/dynamic_modules#compatibility" target="_blank" rel="noopener noreferrer" title="Envoy compatibility" style="font-size: 0.75em; vertical-align: super; text-decoration: none;">?</a>
											)}
```

Comment (sjoukedv) [reply website/src/pages/extensions/[name].astro:100]: Done

nacx (approved):
Thanks!

PR #419, opened 2026-05-11
cli: allow BOE to run custom Envoy binaries

codefromthecrypt:
`--envoy-path` runs any local Envoy binary directly, skipping download and version selection. Useful for testing pre-release or custom-built Envoy.

ext_proc servers now start before Envoy instead of waiting for the admin hook, and `setupDynamicModuleSearchPath` returns the real cleanup function instead of a no-op.

Docker containers default to `BOE_RUN_ID=0` for predictable file paths.

Related: tetratelabs/func-e#493, tetratelabs/func-e#505.

### DIFF
--- a/cli/cmd/healthcheck_test.go
+++ b/cli/cmd/healthcheck_test.go
@@ -20,6 +20,12 @@ import (
 	"github.com/tetratelabs/func-e/experimental/admin"
 )
 
+func TestHealthcheck_Run(t *testing.T) {
+	// In test, pid 1 is not boe, so there's no child with --run-id.
+	err := (&Healthcheck{}).Run(t.Context())
+	require.EqualError(t, err, "timeout waiting for Envoy process: no child with --run-id")
+}
+
 func Test_healthcheck(t *testing.T) {
 	pid := os.Getpid()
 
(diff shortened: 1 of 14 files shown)


cli/cmd/healthcheck_test.go
```diff
@@ -45,46 +45,23 @@ func Test_healthcheck(t *testing.T) {
...
 		// Expect nil error since Run returns nil on context cancellation (documented behavior)
 		require.NoError(t, err)
 
 		require.NoError(t, healthCheckErr)
-		require.Empty(t, log)
+		require.Empty(t, log.String())
 	})
 }
-
-func TestHealthcheck_Run(t *testing.T) {
```
codefromthecrypt, line 68: this test didn't actually do what it said it did, so I deleted it instead. (it didn't fail about context cancel, rather pid 1 stuff)

cli/cmd/run.go
```diff
@@ -280,14 +283,6 @@ func downloadExtensions(ctx context.Context, downloader *extensions.Downloader,
 	return downloaded, nil
 }
 
-// generateRunID generates a unique run identifier based on the current time.
```
codefromthecrypt, line 287: much less complicated to avoid repeating func-e's code

codefromthecrypt (commented):
notes

codefromthecrypt:
similar to https://github.com/envoyproxy/gateway/pull/8958

nacx (approved):
Nice!
And thanks for fixing all the typos :D

cli/cmd/healthcheck_test.go
codefromthecrypt: brought back for coverage

PR #431, opened 2026-05-13
config: set default node identity in generated bootstrap

codefromthecrypt:
Istio [uses ECDS to hot-reload Wasm filters][istio-ecds] on running sidecars without restarting Envoy. In BOE, using [`nativeHttpFilters.after`][native-after], a dynamic module can do the same thing locally: generate [Extension Configuration Discovery Service][ecds] (ECDS) config at init and write it to disk for a native filter to consume (via [`path_config_source`][path-config]). After that point, the module stays out of the request path entirely.

In short, dynamic modules could be used to augment, share code with, or replace control plane ([xDS][xds]) code. This is regardless of if that is about ECDS, [Secret Discovery Service][sds] (SDS) for TLS certs, or any other dynamic config.

If you try this with BOE now, envoy crashes on startup. Envoy requires `node.id` and `node.cluster` for any
xDS config subscription. This simple change to enable this is set `node: {id: "boe", cluster: "boe"}` in `buildFullConfig`. Static configs that don't create subscriptions ignore node entirely, so there's no risk I'm aware of.

[xds]: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol
[path-config]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/config_source.proto#config-core-v3-pathconfigsource
[ecds]: https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/extension#discovery-service
[sds]: https://www.envoyproxy.io/docs/envoy/latest/configuration/security/secret
[istio-ecds]: https://github.com/envoyproxy/envoy/pull/16795
[native-after]: https://github.com/tetratelabs/built-on-envoy/blob/main/extensions/manifest.schema.json

### DIFF
--- a/cli/cmd/testdata/output_full_config.yaml
+++ b/cli/cmd/testdata/output_full_config.yaml
@@ -3,6 +3,9 @@
 # The full text of the Apache license is available in the LICENSE file at
 # the root of the repo.
 
+node:
+  cluster: boe
+  id: boe
 admin:
   address:
     socket_address:
--- a/cli/cmd/testdata/output_full_config_with_cluster.yaml
+++ b/cli/cmd/testdata/output_full_config_with_cluster.yaml
@@ -3,6 +3,9 @@
 # The full text of the Apache license is available in the LICENSE file at
 # the root of the repo.
 
+node:
+  cluster: boe
+  id: boe
 admin:
   address:
     socket_address:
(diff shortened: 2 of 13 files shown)


wbpcode:
But sadly, both Istio and EG haven't provided great support to ECDS. Although the ECDS could provides more smoothy experience (without draining filter chain/listener), but I guess that's because in most cases, the LDS is enough for dynamic updating of filters in practice. :)

Anyway, back to this PR self, that's harmless and LGTM. Thanks for this contribution. :)

wbpcode (approved):

PR #432, opened 2026-05-14
cli: remove codegen dependencies so go run @branch works

codefromthecrypt:
`go run github.com/tetratelabs/built-on-envoy/cli@branch create --type go my-ext` fails because `go generate` does not run during remote module fetches.

Two generated artifacts are behind this: the manifest schema JSON copy and the LibComposerVersion const.

This PR removes both. The schema is now embedded from a root `extensions` package. The composer version is resolved from the OCI registry at `boe create` time, with `--composer-version` to set it explicitly.


nacx:
Thanks!

WDYT about trying to completely remove the codegen step? The embedded manifests were needed initially, but now we publish the manifests [here](https://builtonenvoy.io/extensions.json), and the embedded uses are just a leftover we've not had time to remove.

This is where the embedded manifests are used today:

* This variable holds the embedded manifests: https://github.com/tetratelabs/built-on-envoy/blob/09b5c218cc290e173f4a44945cfb5cf32fcce586/cli/internal/extensions/manifest.go#L222-L223
* It is used in unit tests: We use unit tests to `Validate` the manifests and make sure no invalid manifests are added. This probably can be done in some other way, so we keep the Validation of manifests but without relying on the embedded (code-generated) ones.
* In the `resolveLocalVersions` method, to resolve the Envoy and Composer versions from a manifest against its parent: https://github.com/tetratelabs/built-on-envoy/blob/09b5c218cc290e173f4a44945cfb5cf32fcce586/cli/internal/extensions/manifest.go#L396-L397 We should be able to completely remove this "fallback" when https://github.com/tetratelabs/built-on-envoy/pull/430 is merged.
* It is used in the `ManifestIndex()` method. That is used to generate the public extension index for the website. That is done bu one of the `cli/tools` at build time. The method is also used in the `boe list` unit tests. https://github.com/tetratelabs/built-on-envoy/blob/09b5c218cc290e173f4a44945cfb5cf32fcce586/cli/internal/extensions/manifest.go#L353-L369 We should be able to change the "public manifest list" generation so it does not rely on embedded stuff.
* Finally, the embedded `libcomposer-version.txt` is used in the `boe create` Go template. We should be able to come up with another way of populating that. https://github.com/tetratelabs/built-on-envoy/blob/09b5c218cc290e173f4a44945cfb5cf32fcce586/cli/cmd/templates/create/go/manifest.yaml.tmpl#L2-L10

I think these are all leftover uses of the generated code, and I think we should be able to get rid of those without a lot of trouble, and that would allow us to cleanly use `go run` without having to commit the generated code to Git.

Do you think it is worth tackling the removal of the generated code instead of commiting it to Git?

codefromthecrypt:
@nacx this works and is far more convenient when developing as I can 100pct control boe just like any other go tool like Hugo. how to achieve this, if you have a better idea.. all ears!

codefromthecrypt:
wires crossed.. I didn't see your notes. yep I prefer less checked in code and will rework this in a draft.

codefromthecrypt:
finally ready @nacx I think this is the best I can figure out at the moment for a scaffolding system where there are two versions to track and one isn't in the same go mod

nacx (approved):
This is an amazing cleanup. Thank you very much!

PR #442, opened 2026-05-22
ci: use zig as the C compiler for all composer platforms

codefromthecrypt:
composer-lite only published linux/arm64 and linux/amd64. On macOS,
`boe run` downloads composer-src and builds from source because there is
no darwin/arm64 binary. Docker buildx cannot produce darwin binaries
natively since containers always run Linux.

Replace gcc with zig as the C compiler for all platforms. Zig
cross-compiles to any OS/arch from a single Linux builder, so the same
Dockerfile now produces linux/arm64, linux/amd64, and darwin/arm64
without conditional logic. A `zigcc` wrapper filters linker flags that
zig cannot resolve without a full platform SDK; these resolve at load
time on the target OS.

```console
$ docker buildx build --platform=darwin/arm64     --output type=local,dest=. -f Dockerfile .
$ file libcomposer.so
libcomposer.so: Mach-O 64-bit dynamically linked shared library arm64
```

### DIFF
--- a/extensions/composer/Dockerfile.plugin
+++ b/extensions/composer/Dockerfile.plugin
@@ -51,7 +51,7 @@ RUN ZIG_TARGET=$(echo $TARGETARCH | sed 's/arm64/aarch64/;s/amd64/x86_64/') && \
     case "$TARGETOS" in darwin) ZIG_TARGET="$ZIG_TARGET-macos" ;; *) ZIG_TARGET="$ZIG_TARGET-linux-gnu" ;; esac && \
     CC="zigcc -target $ZIG_TARGET" \
     CGO_ENABLED=1 GOOS=$TARGETOS GOARCH=$TARGETARCH \
-    go build -trimpath -ldflags="-s" -buildmode=plugin ${BUILD_TAG} -o plugin.so ./$EXTENSION_PATH/standalone
+    go build -trimpath -ldflags="-s -w" -buildmode=plugin ${BUILD_TAG} -o plugin.so ./$EXTENSION_PATH/standalone
 
 FROM scratch AS final
 
--- a/extensions/composer/Dockerfile.plugin
+++ b/extensions/composer/Dockerfile.plugin
@@ -7,20 +7,51 @@
 # If you need to customize the build process, consider creating a separate
 # Dockerfile (e.g., Dockerfile.custom) instead of editing this file.
 
-# Build the manager binary
-FROM golang:1.26.3 AS builder
+# Pin the builder to the CI runner platform so cross-OS compilation
+# (e.g. darwin from a Linux runner) works without OS-level emulation.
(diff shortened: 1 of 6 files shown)


nacx (commented):
TIL!

For consistency, should we also update the `Makefile.plugin` so that we also produce `darwin/arm64` images for the individual extensions? This way, `boe run` would fully work for pulling extensions without a local build.

Also, for consistency, I think we should also offer `darwin/arm64` packages for Rust extensions. As far as I've seen, zig can be used as well for Rust cross compilation, and if it works, using the same tool would be great. We can defer this to another PR, but it would be great to have consistency in the platforms we publish packages for in all extensions.

codefromthecrypt:
Extended zig cross-compilation to Go plugins (`Dockerfile.plugin`) and Rust extensions (`Dockerfile.rust`).

Rust darwin/arm64 is still WIP

extensions/composer/Dockerfile.plugin
```diff
@@ -7,20 +7,49 @@
...
+RUN chmod +x /usr/local/bin/zigcc
+
 RUN go mod download all
 
 # BUILD_TAG is used to to include coraza waf build tags, populated by Makefile.plugin via --build-arg
 ARG BUILD_TAG=""
-# Build go plugin
-RUN CGO_ENABLED=1 go build -trimpath -buildmode=plugin ${BUILD_TAG} -o plugin.so ./$EXTENSION_PATH/standalone
+RUN ZIG_TARGET=$(echo $TARGETARCH | sed 's/arm64/aarch64/;s/amd64/x86_64/') && \
+    case "$TARGETOS" in darwin) ZIG_TARGET="$ZIG_TARGET-macos"; EXTRA_LDFLAGS="-s" ;; *) ZIG_TARGET="$ZIG_TARGET-linux-gnu"; EXTRA_LDFLAGS="" ;; esac && \
```
nacx: out of curiosity, why are you adding the `-s` here only for mac? (I see it is not used when compiling the cshared composer).

codefromthecrypt: sorry should have marked draft.. still banging this around until sort out the darwin/arm64

codefromthecrypt: so to add debug symbols on darwin might take work, I stripped regardless ;) if we need debug I can dig more into it lemme know!

nacx: Should we then add `-w` as well to skip debug info completely? Not a big deal, but since we're already stripping stuff this could lead to smaller extension packages.

nacx (approved):

extensions/composer/Dockerfile.plugin
codefromthecrypt: done!

PR #443, opened 2026-05-22
cli: resolve parent manifest from registry for standalone local extensions

codefromthecrypt:
Running a local Go extension outside the built-on-envoy tree fails because
parent manifest resolution only walks the filesystem. Extensions in
standalone repos (like mcp-openapi-boe) never find the parent this way.

Fall back to downloading the parent from the OCI registry when the local
walk returns nothing. This is the same parent manifest that remote
extensions already receive co-packaged in their OCI artifact.

```console
$ cd /tmp && mkdir my-ext && cd my-ext
$ cat manifest.yaml
name: my-ext
parent: composer
type: go
$ boe run --local .   # now resolves parent from registry
```


cli/cmd/run.go
```diff
@@ -400,6 +404,22 @@ func loadLocalManifests(ctx context.Context, logger *slog.Logger, downloader *ex
...
+// resolveParent finds the parent manifest locally, falling back to the registry.
+func resolveParent(ctx context.Context, downloader *extensions.Downloader, m *extensions.Manifest) (*extensions.Manifest, error) {
+	parent, err := extensions.FindLocalParentManifest(m)
+	if err != nil {
+		return nil, err
+	}
+	if parent != nil {
+		return parent, nil
+	}
+	dl, err := downloader.DownloadComposer(ctx, "latest", extensions.ComposerArtifactLite)
```
nacx: This is a bit tricky to do generically (mostly because of the poor definition of "parent" that the current implementation has).

For "out of tree" plugins, the initial idea is that they don't define a "parent" (that was more thought for in-tree plugins), and they explicitly set the `composerVersion` in the manifest. This is how plugins created by `boe create` are generated.

I see value, though, in allowing those to reference the parent, but then I think we need to consider a couple of things:

* Instead of pulling "latest", we should be pulling the `composerVersion` from the current extension manifest (if defined), if not defined, pull the `version` if the current manifest (by convention our individual composer extensions share version with the "parent" composer, but this could not be true for "out of tree" extensions, or fallback to "latest".
* We could check if the "parent" value is "composer", to use the `DownloadComposer` or the generic Download method. Right now we have use case only for "composer" as a parent, so this point is not a big deal and we can leave it for later.

nacx (commented):
Thanks! Left some comments for discussion

cli/cmd/run.go
codefromthecrypt: Thanks for the comments I was not very confident this was the right way but it did trigger the right thinking. Will ponder and post a rev

codefromthecrypt:
Addressed both points, I think, @nacx .

- `resolveParent` now uses `composerVersion` from the manifest (falling back to `version`, then `latest`)
- dispatches to `DownloadExtension` for non-composer parents


PTAL!

cli/cmd/run.go
```diff
@@ -360,8 +361,12 @@ func loadLocalManifests(ctx context.Context, logger *slog.Logger, downloader *ex
 			return nil, fmt.Errorf("%w from %s: %w", errFailedToLoadLocalManifest, path, err)
 		}
 
-		if err := extensions.ResolveLocalVersions(manifest); err != nil {
-			return nil, fmt.Errorf("%w from %s: %w", errFailedToLoadLocalManifest, path, err)
+		if manifest.Type == extensions.TypeGo && manifest.Parent != "" {
```
nacx: Should we just check for hte presence of a parent given that methods it calls are generic?

```suggestion
		if manifest.Parent != "" {
```

codefromthecrypt: good point

nacx (approved):

PR #444, opened 2026-05-22
cli: support dev Envoy version natively via func-e v1.6.0

codefromthecrypt:
Testing unreleased Envoy features with BOE previously meant using
Docker and the distroless-dev image tag, which only works on Linux and
bypasses func-e entirely.

func-e v1.6.0 added native dev builds backed by the same daily
archive-envoy pipeline that publishes released versions. The dev key in
envoy-versions.json carries a commitSha that traces back to the exact
envoyproxy/envoy Docker image it was extracted from.

`--envoy-versions-url` (or `ENVOY_VERSIONS_URL`) passes a custom
versions JSON URL through to func-e. `dev` and `dev-latest` are
accepted as Envoy version strings and bypass the semver compatibility
check against extension manifests.

```console
$ boe run --envoy-version dev-latest --local extensions/example-lua     --envoy-versions-url https://archive.tetratelabs.io/envoy/envoy-versions_debug.json
```


nacx (approved):
This is great. Thank you!

PR #449, opened 2026-05-28
extensions: allow dynamic modules to contain multiple filter types

nacx:
Some compiled Dynamic Modules may contain different filter types. See https://github.com/tetratelabs/built-on-envoy/pull/366 as an example.

This PR changes the filter type to a list, so extensions can declare the types. Then, in `boe run` and `boe gen-config`, the `--filter-type` parameter has been added to decide what type to instantiate when loading the extension.

PR #450, opened 2026-05-28
Use yq for a more robust manifest parsing in Makefiles

nacx:
Builds on top of https://github.com/tetratelabs/built-on-envoy/pull/449

Uses `yq` instead of plain `grep` to parse the extension manifests in the Makefiles for a more robust metadata extraction. It uses `go run` to run YQ so that users don't have to install third-party dependencies.

For the Rust-generated projects (with `boe create`), since they don't have a `go.mod`, the `yq` command will be required on the dev machine. This is not required to build and use this project, but for people scaffolding Rust projects. I think it is a reasonable requirement for them if they want to use the generatedd Makefile.


cli/cmd/templates/create/ext_proc/Makefile.tmpl
```diff
@@ -18,14 +18,16 @@ endif
 # Default data home layout for boe
 BOE_DATA_HOME ?= {{ .DataHome }}
 
+YQ               := go run github.com/mikefarah/yq/v4@v4.53.2
```
wbpcode, line 21: Great idea!

wbpcode (commented):
Thanks for the great contribution. One question, should we also upgrade the script in GitHub workflow to use YQ?

nacx:
>should we also upgrade the script in GitHub workflow to use YQ?

Oh, yes good catch. I'll update in this PR

nacx:
Workflows updated :)

Issue #375, opened 2026-04-17
Add support for Listener filters


ovadiagal:
Similar to https://github.com/tetratelabs/built-on-envoy/issues/370

Add support for Listener Filter Dynamic Modules. 

* [x] Add a way to identify the filter type in the extension manifest. https://github.com/tetratelabs/built-on-envoy/pull/374 
* [x] Update the CLI to generate Listener filter dynamic module configuration accordingly. https://github.com/tetratelabs/built-on-envoy/pull/374
* [x] Update the `boe list` command to include the filter type in the output. https://github.com/tetratelabs/built-on-envoy/pull/372
* [ ] Update the `boe create` command to allow scaffolding listener filters.
* [ ] Update the docs to also include listener filter documentation.
* [x] Include the `filterType` in the OCI annotations for the extension packages. https://github.com/tetratelabs/built-on-envoy/pull/377
* [ ] Include the `filterType` in the extension catalog, potentially allowing to search by filter-type, and showing it in the extension details page.

Issue #370, opened 2026-04-16
Add support for Network Filters


nacx:
Add support for Network Filter Dynamic Modules.

* [x] Add a way to identify the filter type in the extension manifest. https://github.com/tetratelabs/built-on-envoy/pull/368
* [x] Update the CLI to generate Network filter dynamic module configuration accordingly. https://github.com/tetratelabs/built-on-envoy/pull/368
* [x] Update the `boe list` command to include the filter type in the output. https://github.com/tetratelabs/built-on-envoy/pull/372
* [x] Update the `boe create` command to allow scaffolding network filters. https://github.com/tetratelabs/built-on-envoy/pull/373
* [ ] Update the docs to also include network filter documentation.
* [x] Include the `filterType` in the OCI annotations for the extension packages. https://github.com/tetratelabs/built-on-envoy/pull/377
* [ ] Include the `filterType` in the extension catalog, potentially allowing to search by filter-type, and showing it in the extension details page.

PR #452, opened 2026-05-29
rust: add scaffolding for other filter types

nacx:
Part of https://github.com/tetratelabs/built-on-envoy/issues/375 and https://github.com/tetratelabs/built-on-envoy/issues/370

### DIFF
--- a/cli/cmd/templates/create/rust/lib_listener.rs.tmpl
+++ b/cli/cmd/templates/create/rust/lib_listener.rs.tmpl
@@ -86,9 +86,7 @@ impl<ELF: EnvoyListenerFilter> ListenerFilter<ELF> for Filter {
     fn on_close(&mut self, _envoy_filter: &mut ELF) {}
 }
 
-fn init(
-    _server_factory_context_ptr: abi::envoy_dynamic_module_type_server_factory_context_envoy_ptr,
-) -> bool {
+fn init() -> bool {
     true
 }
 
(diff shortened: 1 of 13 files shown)


cli/cmd/templates/create/rust/lib_listener.rs.tmpl
```diff
@@ -0,0 +1,137 @@
...
+    ) -> abi::envoy_dynamic_module_type_on_listener_filter_status {
+        abi::envoy_dynamic_module_type_on_listener_filter_status::Continue
+    }
+
+    fn on_close(&mut self, _envoy_filter: &mut ELF) {}
+}
+
+fn init(
+    _server_factory_context_ptr: abi::envoy_dynamic_module_type_server_factory_context_envoy_ptr,
+) -> bool {
```
nacx: Moving to draft. 
When adding the e2e tests for this scaffolding, I found out that the referenced type didn't exist, and this seems to be an issue in the upstream macro definition.
I've opened https://github.com/envoyproxy/envoy/pull/45368 to attempt to fix it upstream.

nacx:
The fix to the SDK has been merged upstream. I've bumped the version to use that and added the corresponding e2e tests. These tests will take some time, as they'll build 3 Rust extensions. Once this PR has verified the tests, I'll push a commit to make them run only in main, guarding them behind the "skip long-running tests" flag.

nacx:
Tests have passed. I'll open a PR to not run these long-running tests on every PR.

nacx:
@wbpcode, this has been bumped to the Envoy version that has the fixed Rust SDK and works fine.
Actually, I've tried the listener extension with Envoy 1.38, and it works fine. Since the broken part was the "client" side, I guess, once the Rust SDK has been fixed, the listener extension works fine.

Anyway, PTAL!

wbpcode:
cc @nacx I guess this require a main merge?

wbpcode (approved):

PR #471, opened 2026-06-06
fix: golang-loader's image extraction

wbpcode:
Fix golang-loader's image extraction logic.

---

We found this bug until today, in some ways, these means no one use composer-lite for now. 😞


nacx (approved):
So we silently broke it when adding the manifest and config schema to the Docker image :(

PR #473, opened 2026-06-06
composer: support run goplugin-loader explicitly

wbpcode:


cli/e2e/run_test.go
```diff
@@ -320,6 +323,123 @@ func dummy() string {
...
+	)
+	runCmd(t, "", "docker", "push", pluginRef)
+	pluginURL := "oci://" + pluginRef
+
+	// Step 2: build the composer-lite image and extract libcomposer.so into the local cache
+	// at <dataHome>/extensions/dym/composer/<version>/libcomposer.so.
+	runCmd(t, composerDir, "make", "build_image", "COMPOSER_LITE=true")
+	composerImage := fmt.Sprintf("%s/composer-lite:%s-linux-%s", registryAddr, version, runtime.GOARCH)
+	composerCacheDir := filepath.Join(dataHome, "extensions", "dym", "composer", version)
+	extractFileFromImage(t, composerImage, "/libcomposer.so", filepath.Join(composerCacheDir, "libcomposer.so"))
```
nacx, line 386: Should we better push it to the registry as well? This way we'll test that a clean `boe run` can download the plugin and the composer-lite without the user having to have anything locally first.

cli/internal/envoy/extension.go
```diff
@@ -170,9 +170,20 @@ func (d DynamicModuleFilterGenerator) GenerateFilterConfig(manifest *extensions.
 
 	// Use the library name (with underscores) as the dynamic module config name.
 	// This is the identifier Envoy uses to reference the loaded module.
+	//
+	// Bundle-hosted extensions (e.g. the goplugin-loader) are instead loaded
+	// through a shared bundle module (e.g. libcomposer.so) named after the bundle.
+	moduleName := manifest.Name
+	if manifest.Bundle != "" {
+		moduleName = manifest.Bundle
+	}
```
nacx: nit: `extensions.ModuleName(manifest)` to unify the code in a single place?

cli/e2e/run_test.go
wbpcode: SGTM.

nacx (approved):

cli/e2e/run_test.go
nacx: Not a blocker for this PR, we can address in another one if you prefer!

PR #482, opened 2026-06-09
install manifest to the cache

wbpcode:


nacx (approved):
Good catch!
Extension packages also contain the config.schema.json (if any). Should we copy it as well, if it exists? Now we're generating it for the go extensions.
I don't think that would be used today, but for completeness about having the cache with the same contents as the downloaded packages we should probably add that.

wbpcode:
> Good catch! Extension packages also contain the config.schema.json (if any). Should we copy it as well, if it exists? Now we're generating it for the go extensions. I don't think that would be used today, but for completeness about having the cache with the same contents as the downloaded packages we should probably add that.

Good point.

cli/cmd/templates/create/ext_proc/Makefile.tmpl
```diff
@@ -86,6 +86,8 @@ install: build  ## Install the built extproc extension to the local BOE data dir
 	@echo "Installing $(NAME)..."
 	mkdir -p $(BOE_DATA_HOME)/extensions/extproc/$(NAME)/$(VERSION); \
 	cp ./out/ext_proc-server $(BOE_DATA_HOME)/extensions/extproc/$(NAME)/$(VERSION)/ext_proc-server;
+	cp manifest.yaml $(BOE_DATA_HOME)/extensions/extproc/$(NAME)/$(VERSION)/manifest.yaml;
+	cp config.schema.json $(BOE_DATA_HOME)/extensions/extproc/$(NAME)/$(VERSION)/config.schema.json;
```
nacx: File won't exist (yet) for extproc or rust. Can we put the existence check on this in all Makefiles, for consistency, so we don't have to think about it again?

nacx (approved):

PR #483, opened 2026-06-10
merge the local/remote source processing

wbpcode:
- For both remote and local extensions, we use unified function to process the compilation of source code.
- Now, when downloaded the remote extension, we will try the make the artifact manifest and the extension manifest be two different manifests. This is pre change to support general extension bundle.
- existing `parent` is reused, all the full-compiled composer could be used as a general extension bundle.

### DIFF
--- a/cli/cmd/run.go
+++ b/cli/cmd/run.go
@@ -15,7 +15,6 @@ import (
 	"maps"
 	"net"
 	"os"
-	"path/filepath"
 	"runtime"
 	"slices"
 	"strconv"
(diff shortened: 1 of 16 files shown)


cli/internal/extensions/libcomposer.go
```diff
@@ -157,3 +157,44 @@ func BuildLibComposer(logger *slog.Logger, dirs *xdg.Directories, composerSrcPat
 
 	return nil
 }
+
+// BuildComposer builds and installs libcomposer from the composer source tree at composerSrcPath
+// into the local cache, tagging it with the given version.
+func BuildComposer(logger *slog.Logger, dirs *xdg.Directories, composerSrcPath string, version string) error {
```
nacx, line 163: Can we add a unit test for this? We can guard it with the long running test flags if needed. WDYT? We already have tests for the similar `BuildLibComposer` method.

cli/internal/extensions/directories.go
```diff
@@ -18,12 +18,43 @@ func LocalCacheManifest(dirs *xdg.Directories, manifest *Manifest) string {
...
+// artifact, or is empty) it is the manifest at the artifact root; otherwise the artifact is treated
+// as a bundle and its source tree is walked to find the named child's manifest.
+func LocalCacheExtensionManifest(dirs *xdg.Directories, artifactManifest *Manifest, artifactType,
+	extensionName string,
+) (string, error) {
+	var base string
+	if artifactType == ArtifactSource {
+		base = LocalCacheExtensionSourceArtifactDir(dirs, artifactManifest)
+	} else {
+		base = LocalCacheExtensionDir(dirs, artifactManifest)
```
nacx, line 32: Worth having a test that exercises this path?

cli/cmd/run.go
```diff
@@ -587,3 +523,70 @@ func validateComposerCompat(manifests []*extensions.Manifest) error {
...
+		// Build dynamic module (currently supports Rust)
+		if err := extensions.BuildDynamicModule(downloader.Logger, downloader.Dirs, rootManifest, rootPath); err != nil {
+			return err
+		}
+	case extensions.TypeExtProc:
+		fmt.Printf("→ %sBuilding %s...%s\n", internal.ANSIBold, rootManifest.Name, internal.ANSIReset)
+		downloader.Logger.Info("building local ext_proc extension", "name", rootManifest.Name, "version", rootManifest.Version)
+		if err := extensions.BuildExtProcBinary(downloader.Logger, downloader.Dirs, rootManifest, rootPath); err != nil {
+			return err
+		}
```
nacx, line 589: It looks like tests are no longer hitting the type extproc and type rust as they used to. I don't see any significant changes in the existing tests, so I'm not sure why this `handleExtensionSource` method is not exercised anymore in tests, but could you have a look at it?

Previously, we were hitting it for rust and extproc (we were assuming a build error, because we don't wanna run expensive builds in unit tests, but the code path was exercised before, and probably we want to keep that).

nacx (commented):
Thanks! Overall LGTM.
I've added some comments on the places where we don't have coverage that I think we should test (there are others, but they are not relevant). Could you take a look at these test paths?

cli/internal/extensions/libcomposer.go
wbpcode: I initially thought this should be covered by e2e tests, but seems no. the code patch is no triggered by our test.

cli/cmd/run.go
wbpcode: because there no actual path for these extensions and we will return early now because I add a check to that before try to build, I can add a fake path in test to enter previous logic.

nacx (approved):

PR #484, opened 2026-06-10
make bundle a general solution rather than only composer

wbpcode:
code support for bundle extension. Based on previous change, this PR is pretty simple. Now you could run extension in this way:

```
boe run --extension composer/example-go --extension dns/listener
```

Because the `/` is not allowed in the extension name (`"^[a-z][.a-z0-9-_]*$"`), it's pretty great to be used as the separator between bundle and extension name. And easy to understand and use.

If there is a `/`, the first part is bundle (artifact), the second part the specific extension name. If there is no `/`, we will fallback to use the extension name as bundle name.


wbpcode:
funce's latest envoy binary is broken which result in the failure of CI.

cli/e2e/run_test.go
```diff
@@ -148,6 +149,17 @@ func TestRustLocalExtension(t *testing.T) {
...
 
+	// Overwrite the dependency in Cargo.toml to the stable version of Envoy 1.38 to
+	// avoid potential compatibility issues make the CI flaky.
+	// TODO(wbpcode): remove this if we get the dependency more stable.
+	cargoTomlPath := dataDir + "/rust-e2e/Cargo.toml"
+	// #nosec G304
+	cargoToml, err := os.ReadFile(cargoTomlPath)
+	require.NoError(t, err)
+	cargoToml = regexp.MustCompile(`rev = "[0-9a-f]+"`).
+		ReplaceAll(cargoToml, []byte(`rev = "f1dd21b16c244bda00edfb5ffce577e12d0d2ec2"`))
```
nacx, line 160: Let's create an issue to have this tracked

cli/internal/extensions/dynamic_module.go
```diff
@@ -97,6 +103,35 @@ func RustLibNameFromName(name string) string {
 	return strings.ReplaceAll(name, "-", "_")
 }
 
+// copyExtensionManifests copies the main manifest.yaml and any sub-extension
+// manifest.yaml files from srcDir into dstDir. The main manifest is placed at
+// dstDir/manifest.yaml; sub-extension manifests are placed under
+// dstDir/manifests/ preserving their relative directory structure.
+func copyExtensionManifests(srcDir, dstDir string) error {
```
nacx, line 119: not strictly needed, but for consistency in the cache contents, should we copy the config schemas as well?

extensions/composer/Dockerfile
```diff
@@ -20,7 +20,12 @@ ARG BUILD_TAG=""
 # Build go shared library
 RUN CGO_ENABLED=1 go build -trimpath -buildmode=c-shared ${BUILD_TAG} -o libcomposer.so ./main
 
+# Collect all sub-extension manifest.yaml files preserving their directory structure.
```
nacx: We should probably package as well the config schemas for each extension (same comment applies to the rest of Makefiles, etc).

nacx (approved):
Overall LGTM.
Only one comment about packaging the config schemas as well, for consistency in the image artifacts and local cache.

cli/internal/extensions/dynamic_module.go
wbpcode: I didn't see any side effects if we copy that, so will do.

nacx (approved):

PR #500, opened 2026-06-16
composer: add manifest for goplugin-loader

wbpcode:
We added the goplugin-loader support before to allow the users could run the goplugin-loader directly.
In this pull request, we added a manifest to expose it in the UI/document.

Note, when goplugin-loader is ran explicitly, now the full composer will be used. This is deliberate change, because the target of the goplugin-loader support here is to make testing the out-of-tree go plugin easily. And full composer has much higher ratio to be used in practice. I hope it could reflect the actual case.

### DIFF
--- a/extensions/composer/goplugin-loader/manifest.yaml
+++ b/extensions/composer/goplugin-loader/manifest.yaml
@@ -48,10 +48,7 @@ longDescription: |
   - `GOPLUGIN_INSECURE` — set to `true` to allow insecure (HTTP) registry connections
 type: go
 tags:
-  - go
   - plugin-loader
-  - dynamic-module
-  - built-in
 license: Apache-2.0
 examples:
   - title: Load a local Go plugin
--- a/extensions/composer/goplugin-loader/manifest.yaml
+++ b/extensions/composer/goplugin-loader/manifest.yaml
@@ -1,0 +1,77 @@
+# Copyright Built On Envoy
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+# NOTE: The goplugin-loader is always built into the composer (and composer-lite)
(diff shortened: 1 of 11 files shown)


extensions/composer/goplugin-loader/manifest.yaml
```diff
@@ -0,0 +1,77 @@
...
+  ## Runtime environment variables
+
+  - `GOPLUGIN_CACHE_DIR` — directory to cache downloaded plugin binaries
+  - `GOPLUGIN_PULL_SECRET` — credentials for authenticating to the OCI registry
+  - `GOPLUGIN_INSECURE` — set to `true` to allow insecure (HTTP) registry connections
+type: go
+tags:
+  - go
+  - plugin-loader
+  - dynamic-module
```
nacx: [nit] This is probably redundant given the type (and I tthink other extensions don't use this tag).

extensions/composer/goplugin-loader/manifest.yaml
```diff
@@ -0,0 +1,77 @@
...
+  The loader is **always** compiled into the Composer (`libcomposer.so`) and Composer Lite
+  (`libcomposer-lite.so`) dynamic modules. It is **not** a standalone, independently
+  installable or releasable Go plugin: because it is the component that loads other Go
+  plugins, it can never itself be compiled as an independent Go plugin.
+
+  Users reference it directly via `--extension goplugin-loader` or
+  `--extension composer/goplugin-loader`, which drives the loader to fetch and load the
+  configured plugin at runtime.
+
+  ## Configuration
```
nacx, line 35: Given that we now advertise it in the extension catalog, can you add a `config.schema.json` as well? This would also allow the UI to autoconfigure the form to let users use it nicely, and would generate the configuration reference page in the docs.

nacx (approved):

PR #502, opened 2026-06-17
cluster-router: minor adjustments to manifest

nacx:


extensions/composer/cluster-router/manifest.yaml
```diff
@@ -118,24 +118,6 @@ longDescription: |
   flows: client → envoy1 → `peer_envoy2` cluster → envoy2 → `remote_svc` cluster
   → backend. Each Envoy makes its own next-hop decision from its own table.
 
-  ## Configuration Reference
```
nacx, line 121: We now auto-generate reference docs for configs (that are linked from the extension page) based on the config schema, so this is redundant. This PR removes it to have only 1 source of truth for config definition.

extensions/composer/cluster-router/manifest.yaml
```diff
@@ -151,6 +133,7 @@ tags:
   - filter
   - routing
   - mesh
+featured: true
```
nacx, line 136: Let's list it in the landing page :)

wbpcode (approved):

PR #507, opened 2026-06-18
ui: allow running extensions more than once

nacx:
Updates the UI to allow running extensions more than once (for example, for the DNS Gateway extension). It also allows running the same extension with different configs. See recording:

https://github.com/user-attachments/assets/7bf0ede0-7ac2-4cca-bcef-12c1384abeaf


wbpcode:
should we just update the manifest of DNS gateway to reflect it's actual case (a bundle with two extensions)? Then current UI needn't to be updated?

There are two extensions in the dns gateway, and it should contains two (tree, plus the bundle manifest) manifests and two config.schema.json

🤔

nacx:
I think we can have both? The current proposed UI is already valuable for other use cases beyond the DNS extension, so we could merge this, then work on a different PR to refactor the DNS extension into a bundle extension, and also enhance the UI to nicely support that use case as well. WDYT?

wbpcode (approved):

PR #517, opened 2026-06-22
extensions: add e2e test framework

### LINKED ISSUE #85: e2e framework for extension developers
Refactor the internal testing packages to allow extension developers to easily run BOE and create tests for their extensions.

* [x] There should be an easy way to start BOE with the extension to be tested, and have an easy way to add assertions for it.
* [x] The project scaffolding should also create the skeleton for the tests and add the corresponding Makefile targets.
* [x] There should be some documentation

nacx:
Fixes https://github.com/tetratelabs/built-on-envoy/issues/85

* Extracts the `cli/internal/testing` utilities to an internal package at the root so that the `RunEnvoy` helpers can be used by extension e2e tests as well.
* Adds the `extension/tests/e2e` directory where extension e2e tests can be added.
* Refactors the existing WAF integration tests into that folder and use `RunEnvoy` for easier testing.


cli/internal/envoy/config.go
```diff
@@ -600,6 +600,9 @@ func buildHTTPConnectionManager(filters []*hcmv3.HttpFilter, testUpstreamCluster
 				},
 			},
 		},
+		HttpProtocolOptions: &corev3.Http1ProtocolOptions{
+			AcceptHttp_10: true,
+		},
```
nacx, line 605: Some of the WAF FTW tests use HTTP 1.0 requests. I think there is no harm in generating this by default when running extensions locally with BOE, so I just added this instead of over-complicating the UX to make this configurable.

internal/testing/run.go
```diff
@@ -105,7 +105,13 @@ func FreePorts(t *testing.T, count int) []int {
 func RunCLI(t *testing.T, cliBin string, args ...string) *os.Process {
 	// Capture logs, only dump on failure.
 	logDir := t.TempDir()
-	buffers := DumpLogsOnFail(t, logDir, "boe Stdout", "boe Stderr")
+
+	var buffers OutBuffers
+	if teeFile := os.Getenv("TEST_BOE_CLI_OUTPUT_FILE"); teeFile != "" {
```
wbpcode: For these type env variables, should we add a document (or auto register a comment) to explain them like what the Istio did.

internal/testing/logging.go
```diff
@@ -70,11 +70,50 @@ func CaptureOutput(labels ...string) OutBuffers {
 	return buffers
 }
 
+// TeeOutput creates labeled log capture writers that also write to a file as writes happen.
+func TeeOutput(t testing.TB, teeFile string, labels ...string) OutBuffers {
+	f, err := os.Create(filepath.Clean(teeFile))
```
wbpcode, line 75: Seems this `f` never be closed? I guess we need a runtime finalizer for this or the TeeOutput?

wbpcode (commented):
Only minor comments. Thanks for proceeding this.

internal/testing/logging.go
nacx: Done

internal/testing/run.go
nacx: Done. I've refactored all env vars into a single file with descriptions for each one, for better discoverability and reuse. PTAL

wbpcode (approved):

PR #519, opened 2026-06-26
Draft: Add a filter for dynamic latency and fault injection

spockz:
This is a PR to bring my filter https://github.com/spockz/envoy-dynamic-module-dynamic-fault-injection to be structured inside the boe repo using the provided integration testing.

The goal of this module is to be able to take a latency and faults distribution as sampled on production and apply it to an otherwise very perfect server, such as a stub/mock server to achieve a close approximation of the actual downstream behaviour. We use this capability extensively to simulate the behaviour of our mesh in large scenarios with many variables on local machines without clogging up resources limiting us to small networks.

## Todo

- [x]  Use the name `dynamic-fault-injection` consistently everywhere. (Sometimes there is _filter iirc).
- [X]  ~~Decide how to handle that a filter can be both on a listener and on a cluster, whereas on a listener one would typically use `matchers` but when clusters are used per "path" (or more specific) one could forego the path matching.~~ Match stays in, if you want to use it, use it, otherwise leave it out to match anything.
- [X]  Add the `boe` commands to the examples on how to use the filter in conjunction with `boe`.
- [X]  How can we use "proper" yaml instead of an "inline string - parsed as yaml" to configure the plugin.


extensions/composer/dynamic-fault-injection/go.mod
```diff
@@ -0,0 +1,15 @@
...
+	github.com/envoyproxy/envoy/source/extensions/dynamic_modules v0.0.0-20260425100846-90594f45b3ed
+	github.com/stretchr/testify v1.11.1
+	go.uber.org/mock v0.6.0
+	gopkg.in/yaml.v3 v3.0.1
+)
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+)
```
nacx: The `go.mod`, `go.sum` and `Makefile` files need to be removed. Those are created when scaffolding an extension with `boe create`, when extensions live "out-of-tree". 
In-tree Go extensions must not have their own `go.mod` and be part of the `composer` bundle, to make sure all Go plugins can be safely loaded. More details about these constraints here: https://builtonenvoy.io/docs/packaging-prod/#dynamic-modules-and-go

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright Built On Envoy
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+name: dynamic-fault-injection
+version: 0.1.0-dev
```
nacx: The extension will be part of the composer bundle. Declaring the parent will inherit the main properties from it (version, Envoy versions, etc).
```suggestion
parent: composer
```

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - Performance
+author: spockz
+description: A dynamic fault injection filter that simulates latency distributions and error responses.
+longDescription: |
+  An Envoy HTTP filter that injects configurable latency and fault responses based on
+  percentile distributions. Supports per-endpoint configuration with route matching,
+  load-based response behavior with grey zone transitions, and upstream timing awareness
+  to inject only the remaining delay needed to match target distributions.
+type: go
+composerVersion: 0.1.0-dev
```
nacx: Not needed when we reference the `parent` composer.
```suggestion
```

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,28 @@
...
+tags:
+  - go
+  - http
+  - filter
+  - fault-injection
+  - latency
+  - testing
+license: Apache-2.0
+examples: []
```
nacx: Examples will be needed, how to `boe run` this extension so users can easily copy and paste them.

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,28 @@
...
+categories:
+  - Traffic Control
+  - Performance
+author: spockz
+description: A dynamic fault injection filter that simulates latency distributions and error responses.
+longDescription: |
+  An Envoy HTTP filter that injects configurable latency and fault responses based on
+  percentile distributions. Supports per-endpoint configuration with route matching,
+  load-based response behavior with grey zone transitions, and upstream timing awareness
+  to inject only the remaining delay needed to match target distributions.
```
nacx: This manifest is used to render the extension webpage in the extension catalog. The long description accepts markdown. Could you elaborate a bit more and add the examples so that the page is nicely rendered on the site? You can test the rendering locally with:
```
make -C cli gen
cd website
npm install
npm run dev
```

extensions/composer/dynamic-fault-injection/internal/fault/config.go
```diff
@@ -0,0 +1,218 @@
...
+// Accepts both YAML and JSON input. When using google.protobuf.Struct as the
+// filter_config type in Envoy, the config is received as JSON (which is valid YAML).
+func ParseConfig(data []byte) (*FilterConfig, error) {
+	var cfg FilterConfig
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("failed to parse filter config: %w", err)
+	}
+
+	// Validate endpoints.
+	for i, ep := range cfg.Endpoints {
```
nacx, line 73: For validations, can you accumulate all errors on an error slice, then return all of them with `errors.Join()`? This goes a long way in UX when a user gets everything that is wrong on the config instead of errors 1 by 1. Apply this to this method and all other validation methods.

nacx (commented):
Thanks!

The extension looks really nice!
I've done a first pass on the main structure, plugin code, and e2e tests (I've not focused on the business logic of sampling distributions, etc).

Overall looks good. I've left a few comments that will hopefully help the CI jobs pass.

BTW, before committing, run: `make -C cli gen`.
This will auto-generate the website stuff for your extension, as well as the UI needed bits.

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: I did the first fixed. Will look at the examples and sites later. I'm wary of running NPM on my main machine so I will sandbox it. :)

spockz:
Should my plugin also come with a CodeOwners file with me in it?

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: The examples for the configuration are there. Now I need to add the examples for the `boe run` cases. I copied this from my existing README for completeness sake.

website/astro.config.mjs
```diff
@@ -5,6 +5,9 @@ import mermaid from "astro-mermaid";
 
 // https://astro.build/config
 export default defineConfig({
+	server: {
```
spockz: Not happy with this `0.0.0.0`. On my machine without this comment `Astro` conjured up all kinds of IP addresses vaguely similar to the IPs of my network interfaces but none of them matching causing it to fail to bind to a port. With the `0.0.0.0` it binds to all available interfaces instead of just local. But when using `127.0.0.1` it only binds to the IPv4 interfaces. Let me know whether that is fine.

extensions/composer/dynamic-fault-injection/README.md
```diff
@@ -0,0 +1,329 @@
...
+          upstream_http_filters:
+            - name: dynamic_modules/latency_fault
+              typed_config:
+                "@type": type.googleapis.com/envoy.extensions.filters.http.dynamic_modules.v3.DynamicModuleFilter
+                dynamic_module_config:
+                  name: latency_fault_module
+                  do_not_close: true
+                filter_name: latency_fault
+                filter_config:
+                  "@type": type.googleapis.com/google.protobuf.Struct
```
spockz, line 223: This is an old value, I will update it. It has to be a string to be able to parse it as a YAML. 

It will look like (by memory):

```yaml
filter_config:
  "@type": type.googleapis.com/google.protobuf.StringValue
   value: |-
     endpoints:
     ...
```

The effect is that the user misses out of code completion and syntax highlighting. Not sure what the better way is.

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,23 @@
...
...mic-fault-injection\" make build-go\n```\n\nThis produces `liblatency_fault_module.so` which can be loaded by Envoy.\n\n### Unit Tests\n\n```bash\nEXTENSION_PATH=\"composer/dynamic-fault-injection\" make test-go\n```\n\n### Integration Tests\n\n```bash\ncd \"built-on-envoy/extensions/tests/e2e\"\nGO_TEST_ARGS=\"-run TestDistributionDelay\" make test\n```\n\nThis builds the module, starts an Envoy instance with the filter configured, and runs tests against it using the boe built-in test harness.\n\n## Comparison with Envoy's Built-in Fault Filter\n\n| Feature | Built-in Fault Filter | This Module |\n|---------|----------------------|-------------|\n| Fixed delay | ✅ | ✅ (flat distribution) |\n| Percentile distributions | ❌ | ✅ |\n| Per-status-code distributions | ❌ | ✅ |\n| Load-based degradation | ❌ | ✅ |\n| Grey zone transitions | ❌ | ✅ |\n| HTTP abort | ✅ | ✅ |\n| gRPC abort | ✅ | \U0001F6A7 (planned) |\n| Header-controlled faults | ✅ | ❌ (route-based instead) |\n| Response rate limiting | ✅ | ❌ |\n| Per-route configuration | Via per-route config | Built-in route matching |\n| Runtime configuration | ✅ | ❌ |\n| Exact distribution over N requests | ❌ | ✅ (stateful distribution) |"
```
nacx: Can you change this to a multiline string (like in the other yaml manifests) so it is easy to read and maintain?

nacx: Also:
* Remove the Config schema and config fields table from the manifest. The rendered page will already link to the config schema reference, so that is redundant and brings noise to the extension page.
* Replace the Envoy Config example with the correponding `boe run` in the example section. Reason for this is that there is the `boe gen-config` command users can use to get the Envoy config for their extension if they want the full Envoy config, but we keep the extension config siumple, focused just o the extension.
* Remove the Development section from the manifest.

extensions/composer/dynamic-fault-injection/README.md
nacx: Afree on the code completion, but the StringValue is the right way of configuring it.

extensions/composer/dynamic-fault-injection/update-long-description-from-readme.sh
```diff
@@ -0,0 +1,30 @@
...
+  echo "error: manifest not found at $MANIFEST_PATH" >&2
+  exit 1
+fi
+
+README_CONTENT=$(cat "$README_PATH")
+
+# Support either manifest key style.
+README_CONTENT="$README_CONTENT" yq eval -i '.longDescription |= strenv(README_CONTENT)' "$MANIFEST_PATH"
+
+echo "Updated long description in $MANIFEST_PATH from $README_PATH"
```
nacx, line 35: Let's differentiate between the manifest and the readme.

* Manifest: Information for extension Users. It should highlight: what the extension does, its value-add, and how to use it, with examples.
* README: more focused on developers who want to play with it and hack around.

Let's keep this script that syncs the readme and properly scope each for the intended audience.

website/astro.config.mjs
nacx: Astro's default is localhost/loopback-only precisely so a dev server isn't network-exposed by default. Could this be more of an issue on your setup? 

Instead of hardcoding this in the Astro config, can you try running with  `npm run dev -- --host` (or `--host 0.0.0.0`) on your machine and see if it works for you?

nacx:
>Should my plugin also come with a CodeOwners file with me in it?

We already have codeowners. Happy to add you as a codeowner for this extension once it is merged.

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
```diff
@@ -0,0 +1,349 @@
...
+// Sample returns the next pre-computed duration, reshuffling when the cycle completes.
+func (spd *StatefulProbabilityDistribution) Sample() time.Duration {
+	if spd.index >= len(spd.values) {
+		cryptoShuffle(len(spd.values), func(i, j int) {
+			spd.values[i], spd.values[j] = spd.values[j], spd.values[i]
+		})
+		spd.index = 0
+	}
+	val := spd.values[spd.index]
+	spd.index++
```
nacx, line 132: If I followed the code correctly, this could cause race conditions:

`buildFilterFactory` builds one `*latencyFaultFilterFactory` holding `endpoints[].distribution`. The `Create(handle)` returns a new per-request `latencyFaultFilter` that references the same shared factory. On each request, `OnRequestHeaders` calls:

```
ep.distribution.Sample() → ResponseDistribution.Sample() → StatefulProbabilityDistribution.Sample()
```

which mutates shared state without a lock.

extensions/composer/dynamic-fault-injection/plugin.go
```diff
@@ -0,0 +1,273 @@
...
+		if !fault.MatchRoute(ep.match, path, adapter) {
+			continue
+		}
+
+		// Matched endpoint — sample a response.
+		if ep.distribution != nil {
+			f.sample = ep.distribution.Sample()
+			f.matched = true
+		} else if ep.loadBased != nil {
+			// TODO: Feed actual RPS when tracking is implemented.
```
nacx, line 88: IIUC, there is no mention in the README/manifest that this is not yet implemented. Should we document it as such there?

extensions/composer/dynamic-fault-injection/plugin.go
```diff
@@ -0,0 +1,273 @@
...
+		ep := &f.factory.endpoints[i]
+		if !fault.MatchRoute(ep.match, path, adapter) {
+			continue
+		}
+
+		// Matched endpoint — sample a response.
+		if ep.distribution != nil {
+			f.sample = ep.distribution.Sample()
+			f.matched = true
+		} else if ep.loadBased != nil {
```
nacx, line 87: `loadBased` settings are ignored if `distribution` is set, but the configuration allows setting both. Should this be validated at config laod time?

extensions/tests/e2e/dynamic_load_filter_test.go
```diff
@@ -0,0 +1,520 @@
...
+		}
+		if d > actualMaximalDelay {
+			actualMaximalDelay = d
+		}
+	}
+	numRequests := len(durations)
+	avgDelay := totalDelay / time.Duration(numRequests)
+	t.Logf("average request time: %v", avgDelay)
+
+	require.GreaterOrEqual(t, actualMinimalDelay.Milliseconds(), int64(expectedMinDelay),
```
nacx: When comparing, probably you should use `expectedMinDelay.Milliseconds()` as well to make sure you're comparing the right units.

extensions/tests/e2e/dynamic_load_filter_test.go
```diff
@@ -0,0 +1,520 @@
...
+		requireMinimalMaximalAndAverageDurations(t,
+			[]time.Duration{elapsed},
+			upstreamTime, 20*time.Millisecond,
+			time.Duration(minimalDelay)*time.Millisecond,
+			time.Duration(maximalDelay)*time.Millisecond,
+			100*time.Millisecond)
+		require.GreaterOrEqual(t, elapsed.Milliseconds(), int64(upstreamDelayMs-10),
+			"elapsed time should be at least the upstream delay (~50ms)")
+		require.LessOrEqual(t, elapsed.Milliseconds(), int64(maximalDelay+100),
+			"elapsed time should be at most p100 + buffer")
```
nacx: Here and in other e2e tests: aren't these assertions already covered by the helper?

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
```diff
@@ -0,0 +1,349 @@
...
+	// Fill remaining slots (tail beyond last percentile).
+	lastDuration := prevDuration
+	remaining := resolution - idx
+	for i := 0; i < remaining; i++ {
+		fraction := float64(i) / float64(remaining)
+		values[idx] = lastDuration + time.Duration(fraction*float64(lastDuration))
+		idx++
+	}
+
+	cryptoShuffle(len(values), func(i, j int) {
```
nacx, line 110: Here and in the others: do we need cryptographically secure random here? This is on the hot data path and could lead to a syscall epr request.

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
```diff
@@ -0,0 +1,349 @@
...
+import (
+	"crypto/rand"
+	"math/big"
+	"sync"
+	"time"
+)
+
+// ProbabilityDistribution samples from a distribution using linear interpolation
+// between percentile boundaries. Stateless — each sample is independent.
+type ProbabilityDistribution struct {
```
nacx, line 20: This type is never used in the filter code, I think. Should we keep it and document it as some future work or jsut remove it?

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: > Can you change this to a multiline string (like in the other yaml manifests) so it is easy to read and maintain?

I had that initially but in my editor (Zed, VSCode) that caused lack of editing and rendering features which is why I added the update script to fill it from the markdown file. I then went into a rabbit hole of making `yq` update the field as a multi-line string and have yet to succeed.

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
spockz: No, it is not strictly required, however, I could not get the linting to accept the non-secure code using an ignore. I'll double check.

website/astro.config.mjs
spockz: Adding `-- --host` works for me to get the same behaviour as binding to `0.0.0.0` via the config. I'm good with this for now.

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
spockz: Good find! Added the lock.

extensions/composer/dynamic-fault-injection/plugin.go
spockz: Added the mention. I plan to fix this before releasing though.

extensions/composer/dynamic-fault-injection/plugin.go
spockz: Changed this to be a validation error on load time when both configurations are present. In our existing solution there was an alternative field toggling between one or the other to allow for easier switching. Not sure whether we want to retain that.

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: Added

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: Done. Also created https://github.com/mikefarah/yq/issues/2797 to investigate why the yq command broke on the presence of some chars.

extensions/composer/dynamic-fault-injection/update-long-description-from-readme.sh
spockz: Also done, see above.

extensions/tests/e2e/dynamic_load_filter_test.go
spockz: Done

ui/schemas/.json
nacx, line 1: Is this file automatically generated or is it some leftover?

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
nacx: Ping

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,243 @@
...
+  - **Resolution-based weighting**: Use `resolution` as both the sampling accuracy and the relative weight for status code selection
+  - **Load-based behavior**: Configure different response profiles based on current RPS with smooth grey-zone transitions
+  - **Grey zone penalties**: Model degradation with spike detection, penalty multipliers, and recovery rates
+  - **Route matching**: Apply different fault configurations via prefix/exact path matching and header matching
+  - **First-match routing**: Endpoints are evaluated in order; first match wins
+  - **Upstream-aware timing**: Measures actual upstream latency and only adds the remaining delay to reach the target — avoids over-delaying when the upstream is naturally slow
+
+  ## Usage
+
+  ### Delay all success responses and introduce a 1% failure rate that returns quickly
```
nacx: Move this entire section under an example. It will render more nicely on the website extension page.
There is no need for a `Usage` section in the main description. We can have a few examples (or just this one).

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,243 @@
...
+  < x-fault-injected-delay: 414.14ms
+  < x-fault-actual-upstream: 896.86µs
+  < x-fault-status: 503
+  < x-fault-added-delay: 413ms
+  ```
+
+  ## Configuration
+  Please find an overview of the possible fields below, followed by an actual example of how to configure the filter as a per cluster http filter.
+
+  ### Configuration Fields
```
nacx, line 142: In general, I'm not a fan of having the config fields here, as it is very easy for them to get outdated.
A reference page is already auto-generated from the configuration schema and automatically linked from the extension page, so this field enumeration is redundant and is just more work to "keep in sync". I'd consider just removing it.

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,243 @@
...
+
+  | Header | Description |
+  |--------|-------------|
+  | `x-fault-injected-delay` | Target duration from the distribution (e.g., "52.3ms") |
+  | `x-fault-actual-upstream` | Actual time the upstream took to respond |
+  | `x-fault-added-delay` | Additional delay injected (target - upstream, only if > 0) |
+  | `x-fault-injected` | Set to "abort" when a non-2xx status was injected |
+  | `x-fault-status` | The status code selected by the distribution |
+
+  ## Comparison with Envoy's Built-in Fault Filter
```
nacx: Perhaps move this up, right after the first "features" section, so that users getting to this extension can quickly understand the delta? First present the features, what they add on upstresm envoy, then detail how it works, then examples and how to run.

extensions/composer/dynamic-fault-injection/update-long-description-from-readme.sh
nacx, line 1: Let's remove this file and the README.catalog, otherswise we have too many things to maintain and keep in sync for very little value.
Let's keep the README if you want for developers getting to the repo, but let's have a clean `manifest.yaml` that is not auto-generated from other markdown files. It is unnecessary overhead and a maintenance burden just to get the manifest right.

extensions/composer/dynamic-fault-injection/run-local.sh
```diff
@@ -0,0 +1,43 @@
...
+                  }
+				}
+			]
+		}
+	]
+}
+END
+)
+echo $CONFIG
+../../../cli/out/boe* run \
```
nacx, line 38: What happens with the `*` if someone has binaries for many platforms in the `out` directory?

I think we should just delete this file and make the examples in the manifest copy-pasteable

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: Right, I added the examples in the main content. Will also add them here.

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: Okay, fair point! I will leave it out in lieu of a general description of the configuration possibilities.

spockz: On top of that, only the top level docs seem to be available in the reference docs:

Image

Any idea how I should be able to troubleshoot this?

extensions/composer/dynamic-fault-injection/README.catalog.md
```diff
@@ -0,0 +1,220 @@
...
+
+The stateful probability distribution is inspired by [distribution-calculator](https://github.com/spockz/distribution-calculator). Given a set of percentiles, it:
+
+1. Pre-computes exactly `resolution` samples by interpolating between percentile boundaries
+2. Shuffles and serves them in random order
+3. Over a full cycle of `resolution` requests, the actual percentile distribution exactly matches the configured one
+
+### Load-Based Behavior
+
+[!IMPORTANT]
```
spockz, line 111: I'm using a GitHub Markdown flavour here which seems to be just shown verbatim. Is there another mechanism I can use? (Beside copying the markup/css and inlining it. :))

Image

extensions/composer/dynamic-fault-injection/manifest.yaml
nacx: It is possible that not all nesting is captured, and some relationships are not properly generated, as the initial schemas were quite simple.

The code that takes the schema and generates the reference pages lives here in case it needs to be improved (no need to do it in this PR): https://github.com/tetratelabs/built-on-envoy/blob/main/cli/tools/gen-manifest-reference/main.go

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: Done

extensions/composer/dynamic-fault-injection/README.catalog.md
nacx: https://github.com/tetratelabs/built-on-envoy/pull/598 adds support for this

extensions/composer/dynamic-fault-injection/run-local.sh
spockz: You are right. This was just for me to run locally against a local running httpbin server since the online version was having issues.

extensions/composer/dynamic-fault-injection/manifest.yaml
spockz: Alright. Shall I leave the current documentation in for the time being then?

nacx: Yes, let's keep it until the generator is fixed.

extensions/composer/dynamic-fault-injection/manifest.yaml
```diff
@@ -0,0 +1,239 @@
...
+  The stateful probability distribution is inspired by [distribution-calculator](https://github.com/spockz/distribution-calculator). Given a set of percentiles, it:
+
+  1. Pre-computes exactly `resolution` samples by interpolating between percentile boundaries
+  2. Shuffles and serves them in random order
+  3. Over a full cycle of `resolution` requests, the actual percentile distribution exactly matches the configured one
+
+  ### Load-Based Behavior
+
+  [!IMPORTANT]
+  As of yet the load-based behaviour needs to be implemented still.
```
nacx: needs this to be properly rendered.
```suggestion
  > [!IMPORTANT]
  > As of yet the load-based behaviour needs to be implemented still.
```

extensions/composer/dynamic-fault-injection/README.catalog.md
spockz: Fixed with proper syntax.

ui/schemas/.json
nacx, line 1: This file needs to be removed

extensions/composer/dynamic-fault-injection/plugin.go
```diff
@@ -0,0 +1,275 @@
...
+func (f *latencyFaultFilter) OnResponseHeaders(headers shared.HeaderMap, _ bool) shared.HeadersStatus {
+	if !f.matched {
+		return shared.HeadersStatusContinue
+	}
+
+	elapsed := time.Since(f.requestStart)
+	remainingDelay := max(f.sample.Duration-elapsed, 0)
+
+	status := headers.GetOne(":status").ToUnsafeString()
+	fmt.Println(f.sample, status)
```
spockz: This debug leftover should go

extensions/composer/dynamic-fault-injection/README.md
```diff
@@ -0,0 +1,313 @@
...
+< x-envoy-upstream-service-time: 0
+< x-fault-injected-delay: 414.14ms
+< x-fault-actual-upstream: 896.86µs
+< x-fault-status: 503
+< x-fault-added-delay: 413ms
+```
+
+## Configuration
+
+The filter is configured as native YAML in the Envoy config using `google.protobuf.StringValue` as the `filter_config` type.
```
spockz, line 164: Document how a `match` can be defined to match a (virtual) hostname, method, and pathTemplate.

spockz: done

extensions/composer/dynamic-fault-injection/plugin.go
spockz: gone

extensions/composer/dynamic-fault-injection/internal/fault/distribution.go
spockz: This is now being used. The user has the option to select the random variant at the cost of precision or stateful at the cost of memory usage. Memory usage is yet to be quantified.

ui/schemas/.json
spockz: I don't know where it comes from. It appeared when I ran the `make gen` command at one point. Verifying.

spockz: Okay, traced it back. I think it was generated because at some point there was a manifest.yaml which name couldn't be read. Removing.

extensions/composer/dynamic-fault-injection/plugin.go
spockz: Implemented

ui/schemas/.json
spockz: Removed

nacx (approved):
LGTM!
I'm not 100% sure about having all the performance details in the manifest. Perhaps that would be better placed just in the README, but not a blocker for me.

spockz:
> LGTM! I'm not 100% sure about having all the performance details in the manifest. Perhaps that would be better placed just in the README, but not a blocker for me.

Ah yes. I will leave the consideration on which mode to use in the manifest and have the full text in the readme. Is there a nice way to refer to the full readme from the manifest.yaml markdown so it is clickable? (Other than a full GitHub link which would need to point to the source after merging and may become stale etc.)

nacx:
>Is there a nice way to refer to the full readme from the manifest.yaml markdown so it is clickable?

I think the full path to GitHub is fine. Since it is a README it is unlikely to become stale and I think it's good enough.

PR #524, opened 2026-07-01
waf: add metric for transaction duration

nacx:
Adds a histogram metric to track WAF tx duration.


M4tteoP (approved):
LGTM. JFYI: Internally coraza computes the rule-computation duration per phase. So far it is logged in audit logs, but we can expose this upstream in the future if we want to come up with more duration metrics granularity

nacx:
> So far it is logged in audit logs, but we can expose this upstream in the future if we want to come up with more duration metrics granularity

Nice!

wbpcode (approved):

PR #525, opened 2026-07-01
boe: add initial wasm support here

### LINKED ISSUE #12: extensions: WASM extension support
Depends on: https://github.com/tetratelabs/envoy-ecosystem/issues/10

To support WASM extensions, we need to:

* [ ] Implement the WASM filter generator: https://github.com/tetratelabs/envoy-ecosystem/blob/f760e7756c58da3bed83c3d86f112a562df559f7/cli/internal/envoy/extension.go#L59-L63
* [ ] Add a `wasm` element to the [extension manifest spec](https://github.com/tetratelabs/envoy-ecosystem/blob/main/extensions/manifest.schema.json) that allows users to configure properties specific to the WASM extension, such as the location of the `.wasm` file to load or image url.

wbpcode:
The wasm support is there. I personally prefer DYM. But the VM level isolation still a good attribute for specific scenarios.

To close https://github.com/tetratelabs/built-on-envoy/issues/12


extensions/Dockerfile.wasm
```diff
@@ -0,0 +1,32 @@
...
+
+WORKDIR /workspace
+COPY . .
+
+RUN go mod download all
+
+# Delegate the build to the extension's Makefile so the build command lives in a
+# single place. The same pattern works for other languages by swapping the base
+# image and reusing the Makefile's `build` target (e.g. cargo for Rust).
+RUN make build
```
nacx: With this we're mandating every in-tree WASM extension has a Makefile. Isn't it better to be more flexibel and do like we do in extproc, and others, to have the build command here?
Not a big deal for me, just looking for alignment on requirements for in-tree extensions.

nacx (commented):
Overall LGTM!

One thing we need tyo add is the workflow to verify easm extensions, very similar tot he workflow we have for extproc, and make sure it triggers if there are changes to wasm extensions.
This way CI always tests wasm extensions as well and enforces quality and coverage.

extensions/Dockerfile.wasm
wbpcode: 🤔 Then, let me support go only first. And re-consider how to support multiple languages in the future.

nacx (approved):

PR #526, opened 2026-07-01
fix(waf): fail-open in phase 3 and 4 entrances

M4tteoP:
- Adds a fail-open for all the conditions where Envoy returns via the encoding path without hitting the waf first via the decoding one. It was leading to a transaction created just at phase 3 (response headers phase). The CRS in this scenario will return a 403 because no rules' initialization happened. 
- Deprecates ModeResponseOnly. It is very prone to miss configurations and it does not properly work with the CRS. Deprecating until a valid use case shows up.


extensions/composer/waf/waf.go
```diff
@@ -240,7 +242,8 @@ func (p *wafPlugin) OnRequestHeaders(headers shared.HeaderMap, endOfStream bool)
 }
 
 func (p *wafPlugin) OnRequestBody(body shared.BodyBuffer, endOfStream bool) shared.BodyStatus {
-	if p.txContext.IsRuleEngineOff() || p.mode == waf.ModeResponseOnly || p.requestBodyProcessed {
+	//nolint:staticcheck // ModeResponseOnly is deprecated but intentionally still handled for backward compatibility.
+	if p.txContext == nil || p.txContext.IsRuleEngineOff() || p.mode == waf.ModeResponseOnly || p.requestBodyProcessed {
```
nacx: Here and in the other cases, we should probably add a log line here when the `txContext == nil`.

M4tteoP: Added a debug log line for all the nil context checks

nacx (approved):

wbpcode (approved):

PR #551, opened 2026-07-14
generate documents for exension set and link it

wbpcode:
AI assisted, I reviewed the golang and ts part.

---

This is used to enhance the document of extensions. Now the bundle: <bundle-name> will be an link to the extension set's document.

image


nacx:
Nice!

Some extension sets (like the dns-gateway) are "runnable". What do you think about treating them just as normal extensions and include them in the main extensions.json instead of having two different ones?
When showing in the catalog we can make the cards show something that tells it is an extension bundle.

If we treat them as any other extension then other things like the config reference, etc, would already be generated for them, etc.

WDYT? worth trying in another branch (as this one is already valuable in case that approach does not look good?)

wbpcode:
> Some extension sets (like the dns-gateway) are "runnable".

This depends. In my previous thoughts, I think `runnable` extension set is a tech debt (a specific implementation to run multiple extensions within same `.so`, before we have the mature bundle support) rather than an abstraction that should get officially support forever. Now, we have great support to bundle, then the previous way may could be deprecated gradually and we may won't have second `runnable` extension set.

That's why I was hesitate to accept your previous commit at the dns-gateway's manifest PR.

---

But I am always open if we see specific value of the `runnable` extension set that current bundle cannot support well, and want to introduce it as an formal abstraction in the BoE. 🤔 In other word, when will suggest the `runnable` extension set rather than current bundle?

nacx:
Fair enough :)

nacx (approved):

PR #552, opened 2026-07-16
dns-gateway: support IPv6 / AAAA virtual IPs

bplotnick:
## What

Adds first-class IPv6 support to the `dns-gateway` extension set. The gateway and its virtual-IP cache were IPv4-only: `base_ip` parsed as `Ipv4Addr`, allocation arithmetic in `u32`, and the resolver only answered `A` queries with `A` records.

## Changes

- **config**: `base_ip` parses as a generic `IpAddr` (IPv4 or IPv6) via `DomainMatcher::parse_base_ip`; `prefix_len` is validated against the per-family bound (`1..=32` for IPv4, `1..=128` for IPv6).
- **resolver**: the `base_ip` address family selects which query type a domain answers — an IPv4 `base_ip` answers `A` (`AAAA` → NODATA), an IPv6 `base_ip` answers `AAAA` (`A` → NODATA) — and `build_dns_response` emits an `A` or `AAAA` record accordingly.
- **virtual-IP cache**: keyed on `IpAddr`, allocation arithmetic on a `u128` offset so a single code path covers both families; CIDR capacity computed per family.
- **lookup**: the dialed virtual IP parses as a generic `IpAddr` so IPv6 virtual IPs reverse-map to their domain.
- **schema/docs**: `base_ip` accepts IPv4 or IPv6; `prefix_len` max raised to 128.

Existing IPv4 behavior is unchanged. Adds unit tests for IPv6 allocation, `AAAA` responses, and address-family isolation.

## Testing

`cargo test`, `cargo fmt --check`, `cargo clippy --all-targets` all clean (46 tests).


nacx:
Thx!
Can you run `make -C cli gen` so that the site is re-generated with the updated manifest and config schemas?
We should document this step somewhere, as this is a gap in the dev docs

bplotnick:
> Thx! Can you run `make -C cli gen` so that the site is re-generated with the updated manifest and config schemas? We should document this step somewhere, as this is a gap in the dev docs

done!

one thing i didn't mention in the description is that this doesn't currently support dual stacked domains. so you have to pick A or AAAA. not sure if that's something that we should add here

nacx:
Overall LGTM. Thanks!

What would it take to support a dual stack?

nacx (approved):

PR #553, opened 2026-07-16
dns-gateway: bare "*" catch-all matcher + optional domain dynamic metadata

bplotnick:
## What

Two small, independently-useful additions for using the gateway as a universal DNS-takeover layer that defers routing/authorization to downstream filters.

### 1. Bare `*` catch-all matcher

`matches_domain` now treats a pattern of `"*"` as matching every queried domain, and the resolver config loader accepts a `"*"` matcher — skipping the DNS-name parse and the bare-wildcard rejection (which are meant for hostnames) while still validating its `base_ip`/`prefix_len`.

Use it as a low-priority final matcher to intercept all DNS and route/authorize downstream (e.g. `tcp_proxy` route selection on the filter state, or an `ext_authz` check) instead of enumerating every domain. Existing exact and `*.suffix` matching is unchanged; the previous "bare wildcard not allowed" rejection only applied to `*`-as-a-hostname and is now the catch-all.

### 2. Optional domain dynamic metadata on the `lookup` filter

The filter always writes the resolved domain to filter state; some consumers instead read Envoy **dynamic metadata** (e.g. a network `ext_authz` filter with `metadata_context_namespaces` forwards a namespace to the authz service). A new optional `domain_metadata: {namespace, key}` config publishes the resolved domain as a dynamic-metadata string under the operator-chosen namespace/key. Omitted by default, so existing behavior is unchanged.

## Testing

`cargo test`, `cargo fmt --check`, `cargo clippy --all-targets` all clean (45 tests). Adds tests for catch-all matching / config validation and for dynamic-metadata publication (present and absent). Schema + README updated for both.

### DIFF
--- a/extensions/dns-gateway/README.md
+++ b/extensions/dns-gateway/README.md
@@ -81,6 +81,7 @@ See the [extension page](https://builtonenvoy.io/extensions/lookup) for the full
 
 - **Exact**: `"example.com"` — matches only `example.com`
 - **Wildcard**: `"*.aws.com"` — matches one subdomain level (e.g. `api.aws.com`) but not `aws.com` itself or nested subdomains like `sub.api.aws.com`
+- **Catch-all**: `"*"` — matches every queried domain, minting a virtual IP for each. Use it as a low-priority final matcher to intercept all DNS and defer routing/authorization to downstream filters (e.g. `tcp_proxy` route selection on the filter state, or an `ext_authz` check) instead of enumerating every domain
 
 ## Configuration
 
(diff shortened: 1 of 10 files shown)


nacx (approved):
LGTM!

Can you run `make -C cli gen`? We need to document this...

The [schema reference docs](https://builtonenvoy.io/docs/reference/extensions/) are automatically generated from the config schemas, so every time those change we need to run the codegen to generate the site ref docs.

bplotnick:
> LGTM!
> 
> Can you run `make -C cli gen`? We need to document this...
> 
> The [schema reference docs](https://builtonenvoy.io/docs/reference/extensions/) are automatically generated from the config schemas, so every time those change we need to run the codegen to generate the site ref docs.

done! btw, this might conflict with https://github.com/tetratelabs/built-on-envoy/pull/552, so i'll fix that up when this merges. I wanted to split each improvement out for ease of review

PR #554, opened 2026-07-16
dns-gateway: bound and rate-limit the virtual-IP cache (fail-closed)

bplotnick:
> **Stacked on #552 (IPv6 support).** This branch builds on that one, so the diff currently shows both commits — please review the second commit ("bound and rate-limit the virtual-IP cache") here, and review/merge #552 first. Once #552 merges, GitHub will narrow this diff to just the hardening commit.

## What

Virtual-IP allocation is driven by untrusted DNS queries, so an unbounded cache is a memory-exhaustion vector: a flood of unique FQDNs grows it without limit. This makes the shared cache defensive and fail-closed.

## Changes

- **Max-entries cap**: once `max_entries` live mappings exist, new domains are refused (`allocate` returns `None` → the resolver mints nothing and, with `fail_open=false`, returns NODATA). An idle sweep is attempted before giving up.
- **Idle-TTL eviction**: mappings unused for `idle_ttl` are reclaimed, opportunistically when the cache is full and on demand via `evict_idle()`. The paired reverse mapping is dropped too, so a reclaimed IP is never handed to a different domain while live.
- **Per-allocation rate limit**: at most `max_allocs_per_window` *new* allocations per fixed `rate_window`; excess is refused. Cache hits for known domains are never rate limited and refresh the entry's idle timer.
- **`flush()`**: clears all mappings, offset counters, and rate state so a recycled virtual IP can never resolve to a stale FQDN after a config generation bump.

Limits are configurable via `CacheLimits` and default to 65536 entries / 1h idle / 1000 allocations per second; the defaults need no configuration. Timekeeping uses a monotonic process clock; `allocate_at`/`evict_idle_at` take an explicit timestamp for deterministic tests.

## Testing

`cargo test`, `cargo fmt --check`, `cargo clippy --all-targets` all clean (55 tests). Adds tests for the cap, idle eviction, reclaim-then-admit, rate limiting, and flush.


nacx (approved):
LGTM. THanks!
Can you rebase the PR?

PR #555, opened 2026-07-16
dns-gateway: dual-stack — one domain can answer both A and AAAA

bplotnick:
> **Stacked on #552 (IPv6 support).** This branch builds on #552, so the diff shows its commits too — please review the last commit ("dual-stack — one domain can answer both A and AAAA") here, and merge #552 first. Once #552 merges, GitHub will narrow this diff to just the dual-stack commit.

## What

#552 made each domain answer a single address family (an IPv4 `base_ip` serves `A`, an IPv6 `base_ip` serves `AAAA`; the other query type returns NODATA). This lets a single domain be served **dual-stack** — both `A` and `AAAA`.

Configure it by listing the domain twice, once per family — **no schema change** (the `domains` list already permits repeated domains; this only relaxes the implicit one-matcher-per-domain assumption):

```json
"domains": [
  {"domain": "*.aws.com", "base_ip": "10.0.0.0", "prefix_len": 24},
  {"domain": "*.aws.com", "base_ip": "fd00::",   "prefix_len": 64}
]
```

## Changes

- **resolver**: matcher selection is now family-aware. A domain is "intercepted" if any matcher's pattern matches it (non-intercepted → pass through). For an intercepted domain we pick a matcher of the *queried* family; if none exists we return NODATA — **never pass-through** — preserving the no-leak property (a client can't resolve a real address of the other family and bypass the gateway). Non-A/AAAA query types on an intercepted domain still return NODATA.
- **virtual-IP cache**: the dedup key becomes `(domain, address family)` instead of domain alone, so one FQDN can hold both an `A` and an `AAAA` virtual IP. Repeat queries within a family still reuse the existing allocation; the reverse IP→domain map is unchanged (virtual IPs are unique across families).

## Compatibility

Backwards compatible — existing single-family configs behave exactly as before; `AAAA` on a v4-only domain still returns NODATA. Dual-stack only kicks in when a second, other-family matcher is added for the domain.

## Testing

`cargo test`, `cargo fmt --check`, `cargo clippy --all-targets` all clean (54 tests). Adds resolver tests (dual-stack answers both families; single-family still NODATAs the other family) and a cache test (one domain holds both families, deduped per family). Docs regenerated via `make -C cli gen`.


bplotnick:
Good catch on the precedence regression — fixed in 5218d97.

The family-aware selection now determines the *owning* matcher by pattern first (first-match by config order, as before dual-stack), then pairs address families **only among entries sharing that exact pattern**. If the winning pattern has no entry of the queried family it returns NODATA (fail closed) instead of falling through to a later, broader matcher — so your IPv4-exact-then-IPv6-`*` example now correctly NODATAs the AAAA query rather than answering from the catch-all. Added a regression test covering exactly that, and documented the same-pattern pairing + first-match precedence in the README and resolver manifest.

extensions/dns-gateway/README.md
```diff
@@ -83,6 +85,25 @@ See the [extension page](https://builtonenvoy.io/extensions/lookup) for the full
...
 
+Each matcher answers a single address family (`A` for an IPv4 `base_ip`, `AAAA` for an IPv6
+`base_ip`); the other query type returns NODATA. To serve a domain **dual-stack** (both `A` and
+`AAAA`), list it twice **with the same pattern** — once with an IPv4 `base_ip` and once with an
+IPv6 `base_ip`:
+
+```json
+"domains": [
+  {"domain": "*.aws.com", "base_ip": "10.0.0.0", "prefix_len": 24},
+  {"domain": "*.aws.com", "base_ip": "fd00::",   "prefix_len": 64}
```
nacx, line 96: [nit] metadata is configured per-matcher, and this could lead to different metadata for the different families of the same addresses. Since this can have an unexpected impact on downstream filters that consume this, is it worth documenting or adding a note?

extensions/dns-gateway/src/cache_lookup.rs
```diff
@@ -87,7 +87,8 @@ impl<ENF: EnvoyNetworkFilter> NetworkFilter<ENF> for CacheLookupFilter {
         let (ip_str, port) = envoy_filter.get_local_address();
         envoy_log_debug!("New connection, local_address={}:{}", ip_str, port);
 
-        let ip: Ipv4Addr = match ip_str.parse() {
+        // Parse as a generic IpAddr so both IPv4 and IPv6 virtual IPs resolve.
+        let ip: IpAddr = match ip_str.parse() {
             Ok(ip) => ip,
```
nacx, line 92: The cache is now keyed on the exact `IpAddr` variant the resolver allocated (based on what `ip_add` returns) for a v4 range. On a dual-stack listener, Envoy can report an IPv4 connection's local address as `::ffff:10.0.0.5`, which parses to `IpAddr::V6(...)`, resulting in a different map key. With this, the lookup below would miss and `Continue` without setting the filter state, and this could have unexpected effects on downstream filters relying on it.

Would using `to_canonical()` work to overcome this issue? It folds the mapped form back to V4, and is a no-op otherwise:
```
Ok(ip) => ip.to_canonical(),
```

nacx (commented):
Thanks!

extensions/dns-gateway/README.md
bplotnick: yeah. I think i'm going to take this PR a different direction. it's too much of a footgun to have them as separate entries. i'll make it more explicit

nacx:
HI @bplotnick any chance you have cycles to take this to completion? We'd love to have the dual-stack support merged.

PR #556, opened 2026-07-16
fix(waf): `waf_tx_duration` computation, finalize tx early

M4tteoP:
- waf_tx_duration now accumulates time spent in the filter callbacks (real WAF-added latency) instead of the whole request lifetime
- Finalize the transaction (record metrics) at the last phase for the mode — on interruption, or once phase 4 (phase 2 for request-only) completes — instead of at OnStreamComplete. This is needed for long-lived streams whose OnStreamComplete can fire long after inspection ends, which would otherwise leave the transaction pinned and metrics emitted way later.

Note: the PR so far is still running ProcessLogging and closing the tx during OnStreamComplete. ProcessLogging is about phase 5 rules and audit log creation (which might involve write on files) and is expected to be run outside the data path. I see two options here:
- If we agree that it is fine for some scenarios to have that delay on ProcessLogging, we can merge this PR as is
- If not, we should explore some async ways to run it when txDone is set, but not keep the request hanging.


extensions/composer/waf/waf.go
```diff
@@ -440,18 +446,55 @@ func (p *wafPlugin) OnResponseTrailers(_ shared.HeaderMap) shared.TrailersStatus
...
+	}
+}
+
+// trackAndFinalizeTx is deferred at the top of every filter callback. It accumulates
+// the time spent in each callback. This includes the full overhead introduced by the WAF.
+// Also, if the transaction is marked as done, it runs its finalization.
+func (p *wafPlugin) trackAndFinalizeTx(start time.Time) {
+	p.wafElapsed += time.Since(start)
+	if p.txDone {
+		p.transactionDone()
```
wbpcode: could we do the maybeMarkTxDone() at the trackAndFinalizeTx()? So, at other positions, we needn't to ensure `maybeMarkTxDone()` is correct called. And that's say:

```
func (p *wafPlugin) trackAndFinalizeTx(start time.Time) {
    if p.txDone {
        return; // quick return to avoid re-entrance.
    }
	p.wafElapsed += time.Since(start)
    maybeMarkTxDone()
	if p.txDone {
		p.transactionDone()
    }
}
```

wbpcode (approved):
LGTM. Thanks.

nacx (approved):

PR #604, opened 2026-08-24
QoL: Introduce generic targets for check, format, lint, test, and test-coverage for golang 

spockz:
There are actually two improvements:

1. The ability to "just" call `make test` to run the proper tests, regardless of the type of the extension. (Fortunately, this type determination was already in place.
2. The ability to forego having to set the EXTENSION_PATH when calling the makefile from the root directory of an extension.

Maybe a tiny 3rd improvement:

The check of the individual extension now *also* calls the cli/check as there might be something wrong in the manifest.yaml which is otherwise not found, and the extensions need to be committed as well. So makes it easier to check all requirements in one go.

### DIFF
--- a/extensions/Makefile
+++ b/extensions/Makefile
@@ -18,6 +18,19 @@ endif
 BOE_DATA_HOME ?= $(HOME)/.local/share/boe
 
 EXTENSION_PATH ?=
+# Try to automatically detect the EXTENSION_PATH if PWD is set from the shell
+ifneq ($(PWD),)
+	MAKEFILE_DIR = $(realpath $(dir $(firstword $(MAKEFILE_LIST))))
+	CWD          = $(realpath $(PWD))
+	ifneq ($(CWD), $(MAKEFILE_DIR)) # check that we are in a subdirectory and not the extension directory itself
+	ifneq ($(filter $(MAKEFILE_DIR)/%,$(CWD)), )
+	ifneq ($(wildcard $(CWD)/manifest.yaml),) # Check whether manifest.yaml exists by resolves to a file which yields the path or empty string
+		EXTENSION_PATH := $(CWD)
+	endif
+	endif
+	endif
+endif
+
 # Ensure the EXTENSION_PATH always be set
 ifeq ($(EXTENSION_PATH),)
   $(error "EXTENSION_PATH is not set. Please set it to the path of the extension.")
--- a/extensions/Makefile
+++ b/extensions/Makefile
@@ -49,6 +49,8 @@ TIMESTAMP        := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 COMMIT_SHA       := $(shell git rev-parse HEAD 2>/dev/null || echo "unknown")
 AUTHOR           := $(shell $(GO_TOOL) yq '.author' manifest.yaml)
 LICENSE          := Apache-2.0
+HAS_GOLANG       := $(if $(strip $(shell find "$(EXTENSION_PATH)" -type f -name '*.go' -print -quit 2>/dev/null)),true,false)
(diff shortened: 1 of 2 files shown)


extensions/Makefile
```diff
@@ -85,6 +98,54 @@ build-extproc:
 build-wasm:
 	@GOOS=wasip1 GOARCH=wasm CGO_ENABLED=0 go -C $(EXTENSION_PATH) build -buildmode=c-shared -o ./out/plugin.wasm .
 
+.PHONY: check
+check: format lint test-coverage
+	$(MAKE) -C ../cli check
```
nacx, line 103: nit: add a comment that this is done to verify the manifests for the website are generated?

extensions/Makefile
```diff
@@ -85,6 +98,54 @@ build-extproc:
...
+
+.PHONY: format
+format:
+	@if [ "$(TYPE)" = "go" ]; then \
```
nacx, line 107: Using TYPE here, which is read from the manifest, is probably not the right thing to do. There will be `ext_proc` and `wasm` extension types, and those may be written in Go/Rust, so they're subject to those checks. It would be better to detect the language based on the extension code, not the manifest type (manifest type does not indicate the language an extension is written in).

nacx (commented):
Thanks! I've left a couple comments

extensions/Makefile
spockz: I added the comment inside the target definition.

extensions/Makefile
```diff
@@ -178,7 +243,6 @@ push_rust: create_builder
 		--file Dockerfile.rust ./$(EXTENSION_PATH)
 
 .PHONY: push_extproc
-push_extproc:
```
spockz, line 181: There was a duplicate target definition here that `makecheck` complained about. Was it needed for something|?

extensions/Makefile
spockz: Replaced it with `HAS_GOLANG` and `HAS_RUST` variables that are true when finding `.go` or `.rs` files. Now this  has the benefit that mixed projects are also supported. Is this more in line of what you were thinking?

The only caveat is that now when the project has no sources you will get an error that it cannot format/lint/etc your code. Not sure about that one. Please resolve this thread if you agree with the current situation. I already resolved the other comment.

extensions/Makefile
```diff
@@ -18,6 +18,19 @@ endif
...
 
 EXTENSION_PATH ?=
+# Try to automatically detect the EXTENSION_PATH if PWD is set from the shell
+ifneq ($(PWD),)
+	MAKEFILE_DIR = $(realpath $(dir $(firstword $(MAKEFILE_LIST))))
+	CWD          = $(realpath $(PWD))
+	ifneq ($(CWD), $(MAKEFILE_DIR)) # check that we are in a subdirectory and not the extension directory itself
+	ifneq ($(filter $(MAKEFILE_DIR)/%,$(CWD)), )
+	ifneq ($(wildcard $(CWD)/manifest.yaml),) # Check whether manifest.yaml exists by resolves to a file which yields the path or empty string
+		EXTENSION_PATH := $(patsubst $(MAKEFILE_DIR)/%,%,$(CWD))
```
nacx, line 29: Should we preserve the extension path set by the user?
```suggestion
		EXTENSION_PATH ?= $(patsubst $(MAKEFILE_DIR)/%,%,$(CWD))
```

nacx (commented):
Overall LGTM. Just one final comment.

nacx:
Mind updating hte DCO? Then we can merge the PR

spockz:
> Mind updating hte DCO? Then we can merge the PR

Yes... i was afraid that would happen by using the ui for accepting the suggestion. 🤣

nacx (approved):

PR #622, opened 2026-09-10
fix(waf): record waf_tx_duration in microseconds

### LINKED ISSUE #611: composer/waf: waf_tx_duration truncates to whole milliseconds, losing the signal at typical CRS latencies
`metrics.RecordTx` in `extensions/composer/waf/metrics.go` records the accumulated WAF processing time as truncated whole milliseconds:

```go
m.txDuration.Record(h, h.RecordHistogramValue, uint64(elapsed.Milliseconds()))
```

Typical CRS request-phase evaluation is 0.5–3 ms, so most transactions record 0 or 1:

- sub-millisecond transactions are indistinguishable from free (they record literally `0`),
- the mean derived from `waf_tx_duration_sum / waf_tx_duration_count` is biased low by ~0.5 ms — at these magnitudes that is 20–35 % of the true value (we measured 1.41 ms reported vs ~2.4 ms actual on a GET-only replay, 5.83 vs ~6.8 ms on a body-heavy one),
- no median is recoverable at the scale the filter actually costs.

Recording `elapsed.Microseconds()` instead would make Envoy's default histogram buckets (0.5 … 3.6e6) resolve the 0.5–30 ms range cleanly, at no extra cardinality. Happy to send a PR.

Secondary note, related to the same metrics file: `waf_tx_blocked`'s `authority` label is the client-sent `:authority`/Host header. On an internet-facing gateway that is unbounded, scanner-controlled cardinality in Envoy's in-process stat store (we already see raw IPs and reverse-DNS names as label values); stats are never expired. A cap, allowlist hook, or hash option would help — can split this into its own issue if preferred.

Observed with composer v0.10.0 as an Envoy Gateway dynamic module.

norman-zon:
`metrics.RecordTx` recorded the accumulated WAF processing time as truncated whole milliseconds. Typical CRS request-phase evaluation is 0.5–3 ms, so most transactions recorded 0 or 1: sub-millisecond transactions were indistinguishable from free, and the mean derived from `waf_tx_duration_sum / waf_tx_duration_count` was biased low by ~0.5 ms — 20–35 % of the true value at these magnitudes (measurements in #611).

Recording `elapsed.Microseconds()` resolves the 0.5–30 ms range within Envoy's default histogram buckets (0.5 … 3.6e6), at no extra cardinality.

- `waf_tx_duration` now records microseconds; the manifest and the catalog entry are updated to match.
- New unit test pins the unit, including the sub-millisecond case that motivated the issue.

**Note for consumers:** this changes the unit of an existing metric. Dashboards and alerts reading `waf_tx_duration` as milliseconds need to divide by 1000 after upgrading. Happy to add a release-notes line wherever you track those.

The secondary point in #611 — unbounded `authority` label cardinality on `waf_tx_blocked` — is deliberately not part of this PR; I'll split it into its own issue for the opt-out you suggested.

Fixes #611


M4tteoP (commented):
Looks reasonable to me. As written, I would just keep in mind that we should add a release note entry for this change

nacx:
Thanks!

The unit change looks right, but switching to microseconds also changes how the metric lands in Envoy's default histogram buckets, and I think that's worth a note in the manifest's Metrics section.

Envoy records dynamic-module histograms with `Unit::Unspecified` ([abi_impl.cc](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/dynamic_modules/abi_impl.cc#L731)), so the default bucket bounds apply verbatim: `{0.5, 1, 5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000, 300000, 600000, 1800000, 3600000}` ([histogram_impl.cc](https://github.com/envoyproxy/envoy/blob/main/source/common/stats/histogram_impl.cc#L138)).

Read as microseconds, that has two consequences:

1. The first eight bounds (0.5–250 µs) sit below typical CRS evaluation, so the range we actually care about (~0.5–10 ms) gets only five buckets: 500, 1000, 2500, 5000, 10000.
2. The top bound is now 3,600,000 µs = 3.6 s, whereas under milliseconds it was an hour. Samples above it are still stored and still correct in `_sum`/`_count` and in the admin `/stats` quantiles, but PromQL's `histogram_quantile()` interpolates only over `le` bounds, so a p99 that falls past the last finite bucket clamps to 3.6 s and stays pinned there. Unlikely for CRS, but silent enough that it's better documented than discovered.

Both are fixable purely in Envoy config by setting custom buckets. WDYT about adding a note in the manifest and some example showing how to configure that (like in the example below), so that users can take it into consideration?:

```yaml
stats_config:
  histogram_bucket_settings:
  - match:
      safe_regex:
        regex: ".*waf_tx_duration"
    # Sharpens resolution across the useful range and lifts the ceiling to 10s.
    buckets: [100, 250, 500, 750, 1000, 1500, 2500, 5000, 10000, 25000, 100000, 1000000, 10000000]

**Comment by norman-zon** [comment]
Added to the manifest's Metrics section (mirrored in the catalog entry): the unit-less recording, the five useful default bounds across typical CRS latencies, the `histogram_quantile()` clamp at the 3.6 s bound, and your bucket example verbatim.

One practical detail from running this on Envoy Gateway made it into the note: `histogram_bucket_settings` matches the stat name **before** tag extraction, so a `prefix` matcher needs the full `dynamicmodulescustom.waf_tx_duration` name — the regex form in the example sidesteps that.

File: extensions/composer/waf/manifest.yaml [hunk 1]
```diff
@@ -103,7 +103,36 @@ longDescription: |
...
+  - The largest finite bound is 3,600,000 µs (3.6 s). `_sum`/`_count` and the
+    admin-interface quantiles stay correct above it, but PromQL's
+    `histogram_quantile()` clamps any quantile falling past the last finite
+    bound to that bound.
+
+  Both can be tuned in the Envoy bootstrap configuration, without touching the
+  extension. Note that the matcher sees the stat name before tag extraction,
+  i.e. with the `dynamicmodulescustom.` prefix — a `prefix` matcher needs the
+  full `dynamicmodulescustom.waf_tx_duration` name, while a regex sidesteps
+  this:
```
nacx: The regexp has `.*` at the beginning, so it matches every prefix, right? I think the regexp is correct?
One thing we could add is a `$` at the end to make it more precise and better match only that metric, but the prefix is already OK, I think?

norman-zon: The regex is correct, and the `.*` does more than tolerate a prefix: `safe_regex` is a full match. `RegexStringMatcher::match` ([matchers.h](https://github.com/envoyproxy/envoy/blob/main/source/common/common/matchers.h#L166)) delegates to `CompiledGoogleReMatcher::match`, which is `RE2::FullMatch` ([regex.h](https://github.com/envoyproxy/envoy/blob/main/source/common/common/regex.h#L35)). The pattern therefore has to cover `dynamicmodulescustom.` as well; a bare `waf_tx_duration` matches nothing, and an unmatched bucket setting is not an error, so that failure is silent.

Same reason `$` would be a no-op: full match anchors both ends already, so `.*waf_tx_duration` won't match `waf_tx_duration_foo` with or without it.

Reworded the note (and the mirrored catalog entry) to say this outright. It only called out the tag-extraction prefix, which is exactly what makes the `.*` read as redundant.

nacx (approved):

